Issue #42, opened 2018-09-06
A developer can toggle development mode on the local_fabric runtime


sstone1:

sstone1:
Delivered under #52!

PR #52, opened 2018-09-06
Allow user to toggle development mode for local_fabric runtime

sstone1:
Add a new command "Toggle Development Mode" that toggles development mode for the automatically created Fabric runtime `local_fabric`. The command will restart the Fabric runtime if it is already running, but not start it if it is already stopped. The tree view will show the "infinity" symbol next to the Fabric runtime if it is in development mode. This contributes to issue #42.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -47,7 +47,8 @@
             "blockchainAPackageExplorer.packageSmartContractProject",
             "blockchainExplorer.startFabricRuntime",
             "blockchainExplorer.stopFabricRuntime",
-            "blockchainExplorer.restartFabricRuntime"
+            "blockchainExplorer.restartFabricRuntime",
+            "blockchainExplorer.toggleFabricRuntimeDevMode"
         ]
     },
     "main": "./out/src/extension",
@@ -179,6 +180,11 @@
                 "command": "blockchainExplorer.restartFabricRuntime",
                 "title": "Restart Fabric Runtime",
                 "category": "Blockchain"
+            },
+            {
+                "command": "blockchainExplorer.toggleFabricRuntimeDevMode",
+                "title": "Toggle Development Mode",
+                "category": "Blockchain"
             }
         ],
         "menus": {
@@ -233,6 +239,14 @@
                 {
                     "command": "blockchainExplorer.restartFabricRuntime",
                     "when": "view == blockchainExplorer && viewItem == blockchain-runtime-item-started"
+                },
+                {
+                    "command": "blockchainExplorer.toggleFabricRuntimeDevMode",
+                    "when": "view == blockchainExplorer && viewItem == blockchain-runtime-item-started"
+                },
+                {
+                    "command": "blockchainExplorer.toggleFabricRuntimeDevMode",
+                    "when": "view == blockchainExplorer && viewItem == blockchain-runtime-item-stopped"
                 }
             ]
         }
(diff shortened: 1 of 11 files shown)


client/package.json
```diff
@@ -233,6 +239,14 @@
                 {
                     "command": "blockchainExplorer.restartFabricRuntime",
                     "when": "view == blockchainExplorer && viewItem == blockchain-runtime-item-started"
+                },
+                {
+                    "command": "blockchainExplorer.toggleFabricRuntimeDevMode",
+                    "when": "view == blockchainExplorer && viewItem == blockchain-runtime-item-started"
```
caroline-church: Can't you just have an or here rather than adding it twice

client/src/commands/toggleFabricRuntimeDevMode.ts
```diff
@@ -0,0 +1,47 @@
...
+import { CommandsUtil } from './commandsUtil';
+import { RuntimeTreeItem } from '../explorer/model/RuntimeTreeItem';
+import { VSCodeOutputAdapter } from '../logging/VSCodeOutputAdapter';
+import { FabricRuntime } from '../fabric/FabricRuntime';
+
+export async function toggleFabricRuntimeDevMode(runtimeTreeItem?: RuntimeTreeItem): Promise<void> {
+    const runtimeManager: FabricRuntimeManager = FabricRuntimeManager.instance();
+    let runtimeName: string;
+    if (!runtimeTreeItem) {
+        runtimeName = await CommandsUtil.showRuntimeQuickPickBox('Enter a name for the runtime');
```
caroline-church: surely they are choosing a runtime not giving it a name

client/src/explorer/model/RuntimeTreeItem.ts
```diff
@@ -21,24 +21,28 @@ import { FabricRuntimeManager } from '../../fabric/FabricRuntimeManager';
...
+
     contextValue = 'blockchain-runtime-item';
 
     private name: string;
     private runtime: FabricRuntime;
     private busyTicker: NodeJS.Timer;
     private busyTicks: number = 0;
 
-    constructor(provider: BlockchainExplorerProvider, public readonly label: string, public readonly connection: any, public readonly collapsableState: vscode.TreeItemCollapsibleState, public readonly command?: vscode.Command) {
+    private constructor(provider: BlockchainExplorerProvider, public readonly label: string, public readonly connection: any, public readonly collapsableState: vscode.TreeItemCollapsibleState, public readonly command?: vscode.Command) {
```
caroline-church: why do you need another function to call the constructor?

client/package.json
sstone1: Nope; already tried that with braces surrounding the or condition; it did not work.

client/src/explorer/model/RuntimeTreeItem.ts
sstone1: Because we need to do async stuff when creating a new RuntimeTreeItem (determining the state of the runtime), and you can't do that in a constructor - that's why the newRuntimeTreeItem function is async and does additional stuff to new. The alternative is to move the additional stuff into a public `initialize` function and then hope everyone remembers to call it; at least this way they have no choice in the matter.

client/src/commands/toggleFabricRuntimeDevMode.ts
sstone1: I've changed the prompt along with all the other instances that you failed to spot in previous reviews 😉

client/package.json
caroline-church: ok you are right it doesn't work :(

Issue #73, opened 2018-09-10
make generate new smart contract hard code contract types


caroline-church:
Generating the list of smart contract types is nice but is just too slow so we should hard code them or make generating them faster

Issue #32, opened 2018-09-06
fix npm version checking for generator_fabric and yo


caroline-church:
npm view generator-fabric version returns the latest available version, npm ls -g generator-fabric | grep . returns the globally installed version, fix this in createSmartContractProjectCommand.ts
add a check for ensuring the user has the latest version of generator-fabric (and yo?), update it if they haven't

PR #90, opened 2018-09-17
Version check for generator-fabric

Jakeeyturner:
This PR will enable the VSCode extension to check for the latest generator-fabric version and will install it if available (solving Issue #32).
It will also speed up smart contract language retrieval by reading the globally installed generator-fabric package.json for contract languages (solving Issue #73)

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -11,7 +11,7 @@
 # limitations under the License.
 out
 server
-node_modules
+/node_modules
 *.vsix
 .nyc_output
 .vscode-test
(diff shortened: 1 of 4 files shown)


client/test/data/lib/node_modules/generator-fabric/package-without-languages.json
```diff
@@ -0,0 +1,47 @@
+{
```
caroline-church: We don't need this file you can just return the stub

PR #107, opened 2018-10-01
Solve generator conflicts

Jakeeyturner:


client/src/commands/createSmartContractProjectCommand.ts
```diff
@@ -101,12 +105,38 @@ export async function createSmartContractProject(): Promise<void> {
...
+            author: 'John Doe',
+            license: 'Apache-2.0'
+        };
+
+        runOptions['skip-install'] = true;
+        await env.run('fabric:contract', runOptions);
+
+        const packageJson: any = ExtensionUtil.getPackageJSON();
+
+        if (packageJson.production === true) {
```
caroline-church: should just use skip-install and add a progress

client/src/commands/UserInputUtil.ts
```diff
@@ -155,4 +155,15 @@ export class UserInputUtil {
 
         return vscode.window.showQuickPick(languages, choseSmartContractLanguageQuickPickOptions);
     }
+
+    public static showGeneratorOptions(prompt: string): Thenable<string | undefined> {
+        const options: Array<string> = ['Overwrite file', 'Skip file', 'Force all files to overwrite', 'Abort generator'];
```
caroline-church: can we make these variables so that you don't need to write them again later

PR #122, opened 2018-10-03
Instantiate logs function and arguments (close #119)

### LINKED ISSUE #119: Log in the console on instantiate
Log what transaction and arguments were used on an instantiate in the console for reference after the command has been run

Jakeeyturner:

### DIFF
--- a/client/src/commands/instantiateCommand.ts
+++ b/client/src/commands/instantiateCommand.ts
@@ -18,6 +18,7 @@ import { FabricConnectionManager } from '../fabric/FabricConnectionManager';
 import { ChannelTreeItem } from '../explorer/model/ChannelTreeItem';
 import { IFabricConnection } from '../fabric/IFabricConnection';
 import { Reporter } from '../util/Reporter';
+import { VSCodeOutputAdapter } from '../logging/VSCodeOutputAdapter';
 
 export async function instantiateSmartContract(channelTreeItem?: ChannelTreeItem): Promise<void> {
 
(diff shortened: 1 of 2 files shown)


client/src/commands/instantiateCommand.ts
```diff
@@ -73,6 +74,10 @@ export async function instantiateSmartContract(channelTreeItem?: ChannelTreeItem
             const fabricClientConnection: IFabricConnection = FabricConnectionManager.instance().getConnection();
             await fabricClientConnection.instantiateChaincode(data.chaincode, data.version, channelName, fcn, args);
 
+            const outputAdapter: VSCodeOutputAdapter = VSCodeOutputAdapter.instance();
+            const message: string = `\nInstantiated with command: '` + fcn + `' and arguments: '` + args + `'`;
```
caroline-church: i think this should go before the `instantiateChaincode` function as even if it fails you still might want to see what you did

caroline-church: you don't need the new line character also if you use backticks you can use ${variable-name} rather than closing and opening quotes

client/src/commands/instantiateCommand.ts
```diff
@@ -71,8 +72,12 @@ export async function instantiateSmartContract(channelTreeItem?: ChannelTreeItem
 
             progress.report({message: 'Instantiating Smart Contract'});
             const fabricClientConnection: IFabricConnection = FabricConnectionManager.instance().getConnection();
-            await fabricClientConnection.instantiateChaincode(data.chaincode, data.version, channelName, fcn, args);
 
+            const outputAdapter: VSCodeOutputAdapter = VSCodeOutputAdapter.instance();
+            const message: string = `Instantiating with command: '${fcn}' and arguments: '${args}'`;
```
caroline-church: that should be function not command

Issue #72, opened 2018-09-10
make the extension work on windows


caroline-church:
- [x] Fix grpc problem - need to `add node_modules\grpc\src\node\extension_binary\node-v57-win32-x64-unknown` to package.json
- [x] npm commands need to run in a shell - this can be done with an option on the spawn command
- [x] Fix create smart contract - this currently runs commands in `bin/sh`
- [x] Fix local fabric - this also currently runs commands in `bin/sh`

caroline-church:
moving out of phase 1 as won't get done in time

jt-nti:
Seems like a local fabric environment is going to be more problematic on Windows. Agreed with @caroline-church to disable those features on Windows for now

jt-nti:
Also need to look at pkcs11 module!

PR #147, opened 2018-10-05
Remove commands from Command Palette on Windows

jt-nti:
Fabric runtime will not work on Windows so don't show the
related commands in the Command Palette

Contibutes to #72

Signed-off-by: James Taylor <jamest@uk.ibm.com>


jt-nti:
@caroline-church it's a built in thing https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts seems to be the list in the docs

PR #151, opened 2018-10-05
[DO NOT MERGE] Error if instantiate with no installed packages (closes #140)

### LINKED ISSUE #140: Attempting to instantiate with no valid (installed) packages to select gets you stuck
The prompt displayed is:

image

Can't find an input it will accept to then give me an error.  As suggested by @heatherlp , could we have something added for info if the list to select from is empty?

e.g. "No options to display.  Packages must first be installed on member peers before instantiating on a channel."

(With the usual rule of "someone should edit all my wording suggestions for brevity"!)

Jakeeyturner:


client/src/commands/UserInputUtil.ts
```diff
@@ -199,6 +199,12 @@ export class UserInputUtil {
             });
         }
 
+        // If there are no installed smart contract packages
+        if (quickPickItems.length === 0) {
+            vscode.window.showErrorMessage('Packages must be installed before they can be instantiated');
```
caroline-church: I think this error message could be improved... something like `There are currently no smart contracts installed on any peers that this channel is joined to. Please install a smart contract before instantiating`

EdProsser: My preference: "No smart contracts are installed on peers in this channel. Install a smart contract before instantiating."

PR #224, opened 2018-10-19
Improved node/npm requirement warning (closes #165)

### LINKED ISSUE #165: Extension fails to load if node not installed
The readme states that you only need node to start a local fabric, however the extension won't activate if node is not installed
```
extension activating
Updating native node modules
Rebuilding native node modules
Could not rebuild native dependencies spawn npm ENOENT
Failed to activate extension see previous messages for reason
```
Maybe this should also output a suggestion that perhaps node is not installed ?

Jakeeyturner:

### DIFF
--- a/client/src/dependencies/DependencyManager.ts
+++ b/client/src/dependencies/DependencyManager.ts
@@ -104,8 +104,9 @@ export class DependencyManager {
                     await CommandUtil.sendCommandWithOutput('npm', ['rebuild', dependency.moduleName, '--target=2.0.0', '--runtime=electron', '--dist-url=https://atom.io/download/electron'], extensionPath, null, outputAdapter, shell);
 
                 } catch (error) {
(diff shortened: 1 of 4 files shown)


client/src/dependencies/DependencyManager.ts
```diff
@@ -105,7 +105,10 @@ export class DependencyManager {
 
                 } catch (error) {
                     outputAdapter.error(`Could not rebuild native dependencies ${error.message}`);
+                    outputAdapter.error('Please ensure that you have node and npm installed');
                     vscode.window.showErrorMessage(`Could not rebuild native dependencies ${error.message}`);
+                    vscode.window.showErrorMessage('Please ensure that you have node and npm installed');
```
caroline-church: Does this look right when you run it?

Jakeeyturner: I just ended up combining the two error messages together to show that they're related.

PR #235, opened 2018-10-22
workaround of pause before opening dialog

mbwhite:
Added a pause before opening the dialogs.  A workaround not solution - but in local testing it appears to be stable. 

Signed-off-by: Matthew B White <whitemat@uk.ibm.com>


mbwhite:
Added some unit tests.

Issue #230, opened 2018-10-22
As a developer i can debug my javascript smart contract


caroline-church:

caroline-church:
done so closing

PR #236, opened 2018-10-23
Debug javascript smart contract

caroline-church:
Added in debugging a javascript smart contract

contributes to https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/230

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -29,12 +29,16 @@
     },
     "categories": [
         "Programming Languages",
+        "Debug",
         "Other"
     ],
     "activationEvents": [
         "*"
     ],
     "actualActivationEvents": {
+        "other": [
+            "onDebug"
+        ],
         "onView": [
             "blockchainExplorer"
         ],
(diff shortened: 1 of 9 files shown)


client/package.json
```diff
@@ -332,7 +336,99 @@
                     "when": "view == blockchainExplorer && viewItem == blockchain-connection-item"
                 }
             ]
-        }
+        },
+        "debuggers": [
+            {
+                "type": "fabric",
```
sstone1: I think we should call this debugger `fabric:node`, `fabric-node`, or something else with `node` in. I think we're going to need separate debuggers for each chaincode language (Go, Java, .NET), each with its own set of configuration properties - trying to combine them all into one might be difficult!

client/src/extension.ts
```diff
@@ -103,6 +104,10 @@ export function registerCommands(context: vscode.ExtensionContext): void {
 
     disposeExtension(context);
 
+    const provider = new FabricDebugConfigurationProvider();
```
heatherlp: where the types at?

client/src/fabric/FabricRuntime.ts
```diff
@@ -161,6 +161,14 @@ export class FabricRuntime extends EventEmitter {
         await this.runtimeRegistry.update(this.runtimeRegistryEntry);
     }
 
+    public async getChaincodeAddress(): Promise<string> {
+        const prefix = this.docker.getContainerPrefix();
```
heatherlp: where the types at?

PR #263, opened 2018-10-29
Improved managed runtime behaviour (close #113)

### LINKED ISSUE #113: Improve local_fabric explorer behaviour
If the user accidentally deletes the local_fabric from the 'fabric.connections' in the user setting, then it is difficult to get the connection details back.
The user would then have to delete the local_fabric from the 'fabric.runtimes', as well as deactivate and reactivate the extension.

We should improve the behaviour of the local_fabric so that users can't accidentally remove the connection.

One suggestion would be to not expose the local_fabric through the user settings, so that a user can never delete or alter the connection/runtime itself.

Jakeeyturner:

### DIFF
--- a/client/src/fabric/FabricRuntimeManager.ts
+++ b/client/src/fabric/FabricRuntimeManager.ts
@@ -72,24 +72,12 @@ export class FabricRuntimeManager {
             await this.runtimeRegistry.add(runtimeRegistryEntry);
             runtimeRegistryAdded = true;
 
-            // Add the Fabric connection to the connection entry.
-            const connectionRegistryEntry: FabricConnectionRegistryEntry = new FabricConnectionRegistryEntry();
-            connectionRegistryEntry.name = name;
-            connectionRegistryEntry.managedRuntime = true;
-            await this.connectionRegistry.add(connectionRegistryEntry);
-
             // Add the Fabric runtime to the internal cache.
             const runtime: FabricRuntime = new FabricRuntime(runtimeRegistryEntry);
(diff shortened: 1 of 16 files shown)


client/src/fabric/FabricRegistry.ts
```diff
@@ -69,7 +69,8 @@ export abstract class FabricRegistry<T extends FabricRegistryEntry> {
         if (!this.exists(name)) {
             throw new Error(`Entry "${name}" in Fabric registry "${this.registryName}" does not exist`);
         }
-        const oldEntries: T[] = this.getEntries();
+        const oldEntries: T[] = this.getEntries().filter((entry: T) => !entry['managedRuntime']); // Get all entries, but filter out managed runtimes
```
caroline-church: FabricRegistry is supposed to be quite abstract but this is adding in implementation specific stuff

client/src/fabric/FabricRuntimeManager.ts
```diff
@@ -73,10 +73,13 @@ export class FabricRuntimeManager {
             runtimeRegistryAdded = true;
 
             // Add the Fabric connection to the connection entry.
-            const connectionRegistryEntry: FabricConnectionRegistryEntry = new FabricConnectionRegistryEntry();
-            connectionRegistryEntry.name = name;
-            connectionRegistryEntry.managedRuntime = true;
-            await this.connectionRegistry.add(connectionRegistryEntry);
+            if (name !== 'local_fabric') {
```
caroline-church: i don't think we need this code now

Issue #232, opened 2018-10-22
As a developer i can debug my smart contract with having to package or install


caroline-church:

caroline-church:
done so closing

PR #287, opened 2018-11-05
Debug without having to package or install

caroline-church:
Automatically packages and installs a smart contract
Also excluded images from the package

contributes to https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/232


client/src/debug/FabricDebugConfigurationProvider.ts
```diff
@@ -18,102 +18,144 @@ import { FabricRuntimeManager } from '../fabric/FabricRuntimeManager';
...
+        }
     }
 
     // tslint:disable-next-line no-empty
     public dispose(): void {
     }
 
+    private getNewVersion(): string {
+        const date: Date = new Date();
+        return `${this.PREFIX}-${date.getHours()}${date.getMinutes()}${date.getDate()}${date.getMonth()}${date.getFullYear()}`;
```
sstone1: This should be YYYYMMDDHHMM so it can be easily sorted in chronological order 😉

sstone1: Also, for consistency suggest you ensure the number of digits (12) for each field.

PR #371, opened 2018-11-23
Instantiate fix

Jakeeyturner:

### DIFF
--- a/client/src/fabric/FabricConnection.ts
+++ b/client/src/fabric/FabricConnection.ts
@@ -142,9 +142,7 @@ export abstract class FabricConnection implements IFabricConnection {
 
         const instantiatedChaincode: Array<any> = await this.getInstantiatedChaincode(channelName);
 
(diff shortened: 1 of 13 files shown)


client/src/fabric/FabricConnection.ts
```diff
@@ -235,6 +233,82 @@ export abstract class FabricConnection implements IFabricConnection {
...
+            fcn: fcn,
+            args: args
+        };
+
+        const network: Network = await this.gateway.getNetwork(channelName);
+        const channel: Client.Channel = network.getChannel();
+
+        const instantiatedChaincode: Array<any> = await this.getInstantiatedChaincode(channelName);
+
+        const foundChaincode: any = instantiatedChaincode.find((chaincode: any) => {
```
caroline-church: can we make the finding chaincode a seperate function as you do the same thing in the instantiate function?

PR #567, opened 2019-02-11
Check for bad success messages (close #502)

### LINKED ISSUE #502: Check for errenous successful error messages
- if user fails to give a package to delete and exits, we print a "deleted package" success command
- same for the connect command

We should do a mass check across our code for these

Jakeeyturner:
I could only find one instance of where success messages were being thrown when they shouldn't have been.


Jakeeyturner:
> can haz test please

Done!

PR #590, opened 2019-02-14
Only show home page on activation once per version (resolves #503)

### LINKED ISSUE #503: Add to workspace after generating contract duplicate homepage
If the Home page is already open, when generating a smart contract project and adding it to the workspace, a duplicate Home page is opened.

We should only be showing one Home page tab.

---

**Potential Fix**
- Pre-activation, close any Home page tabs (based on the title). Then open a new Home page.

sstone1:
Quite a few changes in here:

- Store the current extension version and activation count in the global state; this is persisted across VSCode windows, VSCode restarts, and extension updates
- During activation, determine if the current extension version has changed since the last activation; this allows us to know if the extension has been updated
- If the extension has been updated, act as before - open the home page, and be noisy about the extension being activated (show the output panel)
- If the extension has not been updated, don't open the home page, and don't be noisy about the extension being activated (don't show the output panel)
- Always read the raw package.json file off the disk, to avoid VSCode's often cached one
- Fix the debug unit tests so they work in the USA

This results in the home page being shown once for each version published, unless they disable the home page via user settings. IMO this is enough to show the user once what's new and what's available via the home page.

The extension gets activated many times during normal usage; it gets activated when you open a new folder, or switch from a folder to a workspace, or a workspace to a folder. 

AFAIK there is no state that is only for the current VSCode process, so that we could show the home page again the next time they restart VSCode.

This is a Microsoft approved strategy for doing one time things, see: https://github.com/Microsoft/vscode/issues/34063

### DIFF
--- a/client/src/dependencies/DependencyManager.ts
+++ b/client/src/dependencies/DependencyManager.ts
@@ -39,8 +39,8 @@ export class DependencyManager {
 
     }
 
-    public hasNativeDependenciesInstalled(): boolean {
-        const packageJSON: any = ExtensionUtil.getPackageJSON();
+    public async hasNativeDependenciesInstalled(): Promise<boolean> {
+        const packageJSON: any = await this.getRawPackageJson();
         return packageJSON.activationEvents.length > 1;
     }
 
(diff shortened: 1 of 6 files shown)


sstone1:
> I think we should get some input from @EdMoffatt regarding the UX behaviour on this.
> If we do decide to go with this, I think we should make it very clear on the Home page, that it can be reopened using the Command Palette.

@Jakeeyturner I don't disagree (though we needed to do that anyway regardless of this problem), but I think we can deal with that separately: https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/599

The current behaviour is unacceptable, so we need to stop it ASAP. Today, if you open VSCode, the home page opens, you import FabCar, and then the home page opens again. It's very confusing for users.

PR #604, opened 2019-02-15
Create project browse prompt (close #584)

### LINKED ISSUE #584: I am not prompted what to browse for when creating a new smart contract project
It might be nice to prompt the user to browse to a location to store the new project, in a similar way to the cert/private key prompt when editing a gateway.

I had been side tracked (opening an issue!) part way through the create project task, so it took me a moment to realise what the folder dialog wanted.

Jakeeyturner:

### DIFF
--- a/client/src/commands/UserInputUtil.ts
+++ b/client/src/commands/UserInputUtil.ts
@@ -318,55 +318,6 @@ export class UserInputUtil {
         return vscode.window.showQuickPick(options, quickPickOptions);
     }
 
-    public static async browseEdit(placeHolder: string, connectionName: string, canSelectFolders?: boolean, filters?: any): Promise<string> {
-        const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
-
-        const options: string[] = [this.BROWSE_LABEL, this.EDIT_LABEL];
-        if (placeHolder.includes('certificate') || placeHolder.includes('private key')) {
-            options.pop();
-        }
-        try {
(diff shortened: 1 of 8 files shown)


client/src/commands/UserInputUtil.ts
```diff
@@ -632,6 +632,37 @@ export class UserInputUtil {
 
     }
 
+    public static async saveDirectory(placeHolder: string): Promise<vscode.Uri> {
```
caroline-church: This function is very similar to the browseEdit one, can we create more of a common function?

client/test/commands/userInputUtil.test.ts
```diff
@@ -1257,4 +1257,62 @@ describe('userInputUtil', () => {
...
+                canSelectFolders: true,
+                canSelectMany: false,
+                openLabel: 'Save'
+            });
+
+            should.equal(result, undefined);
+        });
+
+        it('should handle any errors', async () => {
+            const logStub: sinon.SinonStub = mySandBox.stub(VSCodeBlockchainOutputAdapter.instance(), 'log');
```
caroline-church: This should be a spy you don't need to stub it

client/src/commands/UserInputUtil.ts
Jakeeyturner: I've changed the `browseEdit` function and allowed it to be more flexible.

client/test/commands/userInputUtil.test.ts
Jakeeyturner: Done!

PR #654, opened 2019-02-25
Fixed TS tests with JS contract (close #581)

### LINKED ISSUE #581: TypeScript tests fail to run with a JavaScript smart contract
TypeScript tests fail to run for a generated JavaScript smart contract.
This is because there is no `tsconfig.json` created on the project or test generation.
Additionally, the `typescript` and `ts-node` dependencies are not included in a JavaScript smart contract, so this will cause TypeScript tests to fail.

When generating TypeScript tests, we should consider installing these two dependencies and creating a `tsconfig.json` file.

Jakeeyturner:

### DIFF
--- a/client/src/commands/testSmartContractCommand.ts
+++ b/client/src/commands/testSmartContractCommand.ts
@@ -223,12 +223,47 @@ export async function testSmartContract(chaincode?: InstantiatedContractTreeItem
         return;
     }
 
-    // If TypeScript, update JavaScript Test Runner user settings
+    // If TypeScript, update JavaScript Test Runner user settings and create tsconfig.json
     if (testLanguage === 'TypeScript') {
         const runnerArgs: string = vscode.workspace.getConfiguration().get('javascript-test-runner.additionalArgs') as string;
         if (!runnerArgs || !runnerArgs.includes('-r ts-node/register')) {
             // If the user has removed JavaScript Test Runner since generating tests, this update will silently fail
             await vscode.workspace.getConfiguration().update('javascript-test-runner.additionalArgs', '-r ts-node/register', vscode.ConfigurationTarget.Global);
+        }
+
+        // Setup Tsconfig path
+        const tsConfigPath: string = path.join(localSmartContractDirectory, 'tsconfig.json');
+
+        // Tsconfig file content
+        const tsConfigContents: any = {
+            compilerOptions: {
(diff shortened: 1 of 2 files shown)


client/src/commands/testSmartContractCommand.ts
```diff
@@ -223,13 +223,42 @@ export async function testSmartContract(chaincode?: InstantiatedContractTreeItem
...
+                './functionalTests/**/*'
+            ],
+            exclude: [
+                'node_modules'
+            ]
+        };
+
+        // Try to create tsconfig.json file
+        try {
+            await fs.writeJson(tsConfigPath, tsConfigContents, {spaces: '\t'});
```
caroline-church: what happens if one already exists, we should probably only create one if it doesn't exist

client/src/commands/testSmartContractCommand.ts
```diff
@@ -276,8 +305,8 @@ async function installNodeModules(dir: string, language: string): Promise<void>
     let npmInstallOut: string;
 
     if (language === 'TypeScript') {
-        outputAdapter.log(LogType.INFO, 'Installing package dependencies including: fabric-network@1.4.0, fabric-client@1.4.0, @types/mocha');
-        npmInstallOut = await CommandUtil.sendCommandWithProgress('npm install && npm install --save-dev fabric-network@1.4.0 fabric-client@1.4.0 @types/mocha', dir, 'Installing npm and package dependencies in smart contract project');
+        outputAdapter.log(LogType.INFO, 'Installing package dependencies including: fabric-network@1.4.0, fabric-client@1.4.0, @types/mocha, ts-node, typescript');
+        npmInstallOut = await CommandUtil.sendCommandWithProgress('npm install && npm install --save-dev fabric-network@1.4.0 fabric-client@1.4.0 @types/mocha ts-node typescript', dir, 'Installing npm and package dependencies in smart contract project');
```
caroline-church: do we need the first npm install?

Jakeeyturner: First `npm install` has now been removed.

client/src/commands/testSmartContractCommand.ts
Jakeeyturner: We now check if a `tsconfig.json` exists. If it already exists, we warn the user but continue and assume that the config will work.

Issue #722, opened 2019-03-14
Creating a smart contract project is too damn slow


sstone1:
All of the `npm ls` shenanigans we go through have to stop!

PR #723, opened 2019-03-14
Use bundled generator-fabric instead of user installed version (resol…

sstone1:
…ves #722)


caroline-church:
tried this out, its is so much quicker

Issue #724, opened 2019-03-14
After Stopping the Fabric Runtime, the Fabric Gateway panel still tries to reconnect


rthatcher:
When I stop the "Fabric Runtime", the Fabric Gateways panel refreshes and then fails to connect with this error:
![image](https://user-images.githubusercontent.com/29195399/54357091-26491900-4655-11e9-96c8-65ba7c31d14e.png)

Could the Stop function disconnect before stopping?
(v0.3.2)

PR #751, opened 2019-03-22
Disconnect local fabric gateway when stop or tear down

caroline-church:
Disconnect gateway to stop errors from when stopping

closes https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/724


client/test/commands/teardownFabricRuntime.test.ts
```diff
@@ -67,11 +79,28 @@ describe('teardownFabricRuntime', () => {
     });
 
     it('should teardown a Fabric runtime', async () => {
```
Jakeeyturner: What's the difference between this test and the test on line 92?

caroline-church: in the first test the gateway panel isn't connected so doesn't need to disconnect, whereas in the second test it is connected to a local runtime so therefore needs to disconnect

PR #756, opened 2019-03-25
Debug package and install question (close #703)

### LINKED ISSUE #703: Update debug
Currently when you stop debug and start it again it will package and install the contract but this isn't needed unless you change the function you use for instantiate.

We should update the debug logic so that it only packages and installs again if the user requests that otherwise it just starts debugging.

Jakeeyturner:

### DIFF
--- a/client/src/commands/UserInputUtil.ts
+++ b/client/src/commands/UserInputUtil.ts
@@ -185,7 +185,7 @@ export class UserInputUtil {
         return vscode.window.showQuickPick(peerNames, quickPickOptions);
     }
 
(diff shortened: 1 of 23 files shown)


client/src/commands/UserInputUtil.ts
```diff
@@ -715,6 +715,29 @@ export class UserInputUtil {
         return vscode.window.showQuickPick(quickPickItems, quickPickOptions);
     }
 
+    public static async packageAndInstallQuestion(): Promise<IBlockchainQuickPickItem<boolean>> {
+            const quickPickItems: IBlockchainQuickPickItem<boolean>[] = [
```
caroline-church: Can we check the wording with design please?

Jakeeyturner: Done!

PR #808, opened 2019-04-15
updated the readme to give info about debugging

RiehaB:
Signed-off-by: Rieha <rieha.burrell@ibm.com>

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -163,6 +163,8 @@ Debugging your smart contract allows you to run through the smart contract trans
 
 To make iterative changes to your smart contract while debugging, after making your changes click the **restart** button. Restarting debugging means you don't need to instantiate the contract again. Please note, as this stores the smart contract in local memory, for many changes to large smart contracts, you may need to reinstantiate the smart contract.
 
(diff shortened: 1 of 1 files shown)


client/README.md
```diff
@@ -163,6 +163,8 @@ Debugging your smart contract allows you to run through the smart contract trans
 
 To make iterative changes to your smart contract while debugging, after making your changes click the **restart** button. Restarting debugging means you don't need to instantiate the contract again. Please note, as this stores the smart contract in local memory, for many changes to large smart contracts, you may need to reinstantiate the smart contract.
 
+If you stop the fabric runtime, in order to debug you’ll need to rename your smart contract or teardown the fabric runtime.
```
cazfletch: i think you need to say something like `If you stop the fabric runtime after a debug session, you will need to rename your smart contract  or teardown the fabric runtime and restart it to continue the debug session. This is because when fabric starts up again it will automatically create the chaincode container.`

heatherlp:
I've got to do another readMe update for 0.4.0, so I'll put this change in there too

Issue #828, opened 2019-04-18
Packaging still happens if npm run build fails


cazfletch:
If npm run build fails packaging still happens. This shouldn't happen it should give an error

PR #837, opened 2019-04-18
Don't create package if errors

cazfletch:
Now won't create a package if there are errors

closes https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/828


client/src/commands/packageSmartContractCommand.ts
```diff
@@ -129,7 +131,7 @@ export async function packageSmartContract(workspace?: vscode.WorkspaceFolder, o
 
             // Create the package. Need to dynamically load the package class
             // from the Fabric SDK to avoid early native module loading.
-            const {Package} = await import('fabric-client');
+            const { Package } = await import('fabric-client');
```
sstone1: I'm sure that you're always the one deleting these spaces, not adding them!

Issue #776, opened 2019-03-27
As a developer, I can use service discovery with the local Fabric


sstone1:
The local Fabric currently does not enable you to use service discovery. This is because the peer does not currently have an external endpoint address set, and the orderer address in the channel configuration block does not work outside of the Docker network.

This is _okay_, but it means that any sample applications run inside VSCode cannot be configured to use service discovery, and this means that developers must write different code to toggle service discovery on or off depending on if they are using the local Fabric or trying to interact with the SaaS.

Our desire is to see all applications running with service discovery, so we should really be demonstrating this out of the box from the start with the local Fabric working with the samples, and allow you to use those exactly same samples with the SaaS.

heatherlp:
@sstone1 is this done?

PR #841, opened 2019-04-23
Generate local Fabric with correct port configuration (contributes to…

sstone1:
… #776)


sstone1:
> I'm still in the middle of reviewing the code, but I did notice that you're deleting the entire local_wallet on teardown, we don't want that. The user could be using local_wallet for another gateway. So can you update it so that only admin is removed? Like it is today in the master stream.

@heatherlp that one needs some discussion. Identities in the wallet created using the CA are no longer valid after the Fabric is torn down, so surely we should remove those as well?

cazfletch:
How would you know which identities are associated with which fabric though?

sstone1:
> How would you know which identities are associated with which fabric though?

🤷‍♂️ 

However, leaving them there for the user to manually clean up after a teardown operation just seems annoying - I agree that the answer isn't obvious though.

@heatherlp I've updated the PR to only delete the identities that came from the runtime (e.g. the `admin` identity) and leave the other ones in place - can you continue reviewing please?

client/src/explorer/walletExplorer.ts
```diff
@@ -71,48 +71,48 @@ export class BlockchainWalletExplorerProvider implements BlockchainExplorerProvi
...
+                const isAdminIdentity: boolean = adminIdentities.some((adminIdentity: FabricIdentity): boolean => adminIdentity.name === identityName);
+                if (isAdminIdentity) {
+                    // User can't delete this!
+                    tree.push(new AdminIdentityTreeItem(this, identityName, walletName));
+                } else {
+                    tree.push(new IdentityTreeItem(this, identityName, walletName));
+                }
             } else {
-                tree.push(new IdentityTreeItem(this, identityName, walletTreeItem.name));
+                tree.push(new IdentityTreeItem(this, identityName, walletName));
```
heatherlp: (review ongoing) This looks the same as line 112 - do you need both?

sstone1: @heatherlp I suppose not - rewrote the block so that the duplicate line does not occur.

client/src/fabric/FabricRuntime.ts
```diff
@@ -77,6 +82,81 @@ export class FabricRuntime extends EventEmitter {
...
+        });
+
+    }
+
+    public async importWalletsAndIdentities(): Promise<void> {
+
+        // Ensure that all wallets are created and populated with identities.
+        const fabricWalletGenerator: IFabricWalletGenerator = FabricWalletGeneratorFactory.createFabricWalletGenerator();
+        const walletNames: string[] = await this.getWalletNames();
+        for (const walletName of walletNames) {
```
heatherlp: why would there ever be more than one of these for a runtime?

client/src/fabric/FabricRuntime.ts
```diff
@@ -77,6 +82,81 @@ export class FabricRuntime extends EventEmitter {
         return this.state;
     }
 
+    public async create(): Promise<void> {
+
+        // Delete any existing runtime directory, and then recreate it.
+        await fs.remove(this.path);
```
heatherlp: remember yesterday when I had to delete the runtime folder to get stuff to work? shouldn't this code have been run yesterday?

client/src/fabric/FabricRuntime.ts
sstone1: @heatherlp today no, but planning for the future where you might import the "commercial paper advanced" network sample which would have a wallet for each organisation.

client/src/fabric/FabricRuntime.ts
sstone1: @heatherlp it did run, but the generator didn't work (as you hadn't run npm install) so you were left with an empty runtime folder. The check for creation is "does the runtime folder" exist, so it didn't recover from the problem.

client/src/fabric/IFabricWallet.ts
```diff
@@ -17,6 +17,8 @@ export interface IFabricWallet {
 
     importIdentity(certificate: string, privateKey: string, identityName: string, mspid: string): Promise<void>;
 
+    delete(identityName: string): Promise<void>;
```
heatherlp: where's the function associated with this?

sstone1: @heatherlp same place as exists - from the base FileSystemWallet class.

client/test/data/yofn/admin-msp/keystore/5abbeb24b1ac370db9031971a0d34abb4307dc6fc4548a7da9df4ffc4aae1c87_sk
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
heatherlp: why do you need all these files in unit tests?

sstone1: @heatherlp the new code reads files and directories from a network generated by the generator, so I've checked in a generated network to test that code.

heatherlp: rather than stub that out and do this in integration tests?

sstone1: @heatherlp yeah, that's a lot of file system related stubs - and we already do it for other unit tests - there's a lot of stuff under the `client/test/data` directory. All of the operations against the generated network are read only (no writes/deletes/executes).

PR #857, opened 2019-04-26
[Merge after 0.4.0] Dissociate wallet after removal (close #843)

### LINKED ISSUE #843: remove wallet comand ignores wallet association
## Description

On removing a wallet associated with a gateway, this works but the gateway view still shows that a wallet is associated, as does the user settings

## Expected Behavior

A warning or it does it for me?

## Possible Fix

Perhaps we would just want to disassociate the wallet if we've already asked the user whether they want it removed or not?

I think this is sev3 because the dissociate command still works afterwards, so no biggie

## Steps to Reproduce

1. Start with a wallet associated with a gateway
2. Remove the wallet
3. Look at gateways panel

## Existing issues

## Your Environment

* Visual Studio Code version:
* Link to your project:

Jakeeyturner:
After deleting a wallet, any gateways associated to that wallet will have their association removed.


client/src/commands/removeWalletCommand.ts
```diff
@@ -45,6 +47,14 @@ export async function removeWallet(treeItem: WalletTreeItem): Promise<void> {
         await fs.remove(walletRegistryEntry.walletPath);
         await FabricWalletRegistry.instance().delete(walletRegistryEntry.name);
 
+        const gateways: FabricGatewayRegistryEntry[] =  FabricGatewayRegistry.instance().getAll();
```
heatherlp: flag

Jakeeyturner: Just checked and `FabricGatewayRegistry.instance().getAll()` will always return an array (if called from the command, it's possible to be an empty array).

PR #883, opened 2019-04-30
[Merge after v0.4.0] Adding more telemetry for usage statistics

SamRasha:
For GA release. Needs particular attention to the sending of the telemetry events with regards to format of how it is sent (format, properties etc). 

Signed-off-by: SamRasha <samanrasha@hotmail.co.uk>


client/templates/HomeView.ejs
```diff
@@ -446,5 +446,17 @@
             </div>
         </div>
     </div>
+    <script>
+        const vscode = acquireVsCodeApi();
+
+        function openUrl(url) {
+            console.log('the url', url);
```
heatherlp: can you remove this console.log please?

client/src/webview/SampleView.ts
```diff
@@ -70,6 +72,7 @@ export class SampleView extends View {
 
                 // Refresh the webviews html
                 panel.webview.html = await this.getHTMLString();
+                Reporter.instance().sendTelemetryEvent('Sample Opened', {sample: this.sampleName, language: language});
```
Jakeeyturner: I think we'll want to send the `fileName` (FabCar Contract) and `fileType` (contracts/applications) as well

PR #890, opened 2019-05-01
Configuration settings cleanup (close #880)

### LINKED ISSUE #880: Clean up the names of the configuration settings
The extension currently has quite a few configuration settings:

```
    "fabric.runtime": {
        "ports": {
            "orderer": 17050,
            "peerRequest": 17051,
            "peerChaincode": 17052,
            "peerEventHub": 17053,
            "certificateAuthority": 17054,
            "couchDB": 17055,
            "logs": 17056
        },
        "developmentMode": false
    },
    "fabric.gateways": [],
    "fabric.wallets": [],
    "blockchain.ext.directory": "~/.fabric-vscode",
    "extension.home.showOnStartup": true,
    "blockchain.repositories": []
```

These names are mixed in with all other configuration settings from all other extensions. They're not in a namespace for our extension. It would be better if these names were consistent - for example, a single `ibm-blockchain-platform` prefix?

The worst offender is `"extension.home.showOnStartup"` which doesn't mention blockchain, fabric, or IBM. It could be from any extension!

v1.0.0 is our chance to get this right, so we should change it before GA.

Jakeeyturner:

### DIFF
--- a/client/SettingConfigurations.ts
+++ b/client/SettingConfigurations.ts
@@ -1,0 +1,31 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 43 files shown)


Jakeeyturner:
> Can we migrate all the user settings and then delete the unused ones just to make it abit cleaner

Unfortunately not 😞

PR #908, opened 2019-05-03
Enable setting transient data

### LINKED ISSUE #832: Allow submit transaction command to include transient data
Submit transaction command should let you add in transient data

cazfletch:
Enabled setting transient data

closes #832


client/src/commands/submitTransaction.ts
```diff
@@ -90,22 +90,42 @@ export async function submitTransaction(evaluate: boolean, treeItem?: Instantiat
         args = argsString.split(','); // If empty, args will be ['']
     }
 
+    let transientData: { [key: string]: Buffer };
+    try {
+
+        const transientDataString: string = await UserInputUtil.showInputBox('optional: What is the transient data for the transaction, e.g. {"key": "value"}', '{}');
```
heatherlp: what happens if they remove the default value, i.e. leave the field blank and press enter?

cazfletch: if you just press enter it returns '' so no transient data gets set and it send undefined to fabricClientConnection

PR #987, opened 2019-05-21
Stopping users from entering a smart contract name containing a space

SamRasha:
This fix means that although it throws errors, it still creates a folder on the user's machine with the invalid name they entered, not much we can do about that though.
Signed-off-by: SamRasha <samanrasha@hotmail.co.uk>

### DIFF
--- a/client/cucumber/steps/steps.ts
+++ b/client/cucumber/steps/steps.ts
@@ -558,7 +558,7 @@ module.exports = function(): any {
 
         const uri: vscode.Uri = vscode.Uri.file(contractDirectory);
 
-        browseStub.withArgs('Choose the location to save the smart contract', [UserInputUtil.BROWSE_LABEL], {
+        browseStub.withArgs('Choose the location to save the smart contract, please do not include spaces in the folder name', [UserInputUtil.BROWSE_LABEL], {
             canSelectFiles: false,
             canSelectFolders: true,
             canSelectMany: false,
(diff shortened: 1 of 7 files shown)


cazfletch:
Then this is the wrong fix and we should check the name before it creates the folder

SamRasha:
Hi @cazfletch so i've researched different ways to be done during the process of inputting a smart contract name. The only way i think it may? be possible is through the openDialogOptions, in particular the filters. @heatherlp Also has taken a look and says it may not be possible unless somehow it can be done in the filters option. I've tried researching through the docs, but have been unsuccessful. Any ideas?

SamRasha:
I believe it will not work because the filters option is to do with file extensions and not the filename itself that has been chosen by the user. I would also think we may want to refrain from deleting the selected folder after it errors in the case that the folder was pre-existing/user does not want to delete it. I think for now what I can think of is perhaps making the browse information line here 
Screenshot 2019-05-21 at 15 46 59 change from 'Choose the location to save the smart contract' to something along the lines of 'Choose the location to save the smart contract, please do not include spaces in the folder name'

SamRasha:
Added checking and enforcement of correct usage of package names during the package smart contract process. Revamped tests to fit new enforced style.

Stashed the create smart contract enforcement of package names changes because it would require revamping the temp directory process used in the tests. Attempting to finish by tomorrow

Jakeeyturner:
Closing this PR as I've picked up this issue!

Issue #967, opened 2019-05-14
As a first time user I'd like help with installing extension prerequisites


heatherlp:
Notes here on what this could mean:
- People can get too far without having them all in place, or can get really stuck if they install it and can’t compile gRPC. It seems like it’s difficult to resolve the problem if gRPC fails, you fix the problem, it doesn’t automatically resolve in the IDE. You also need to reinstall the extension.
We should also try to see if we can require grpc at startup and assume we need to rebuild if it won’t load

- We also need to support Node 12

- A windows guy just got stuck with a 32 bit VSCode with a 64 bit Node and a 64 bit grpc and there was no hints as to what the problem was

heatherlp:
I've just done a quick test with node 12 and v1.0.0, instantiating and debugging on a newly created smart contract project so I don't think much (if any) work is needed to support it

Jakeeyturner:
[Prerequisites Hi-Fi.zip](https://github.com/IBM-Blockchain/blockchain-vscode-extension/files/3321208/Prerequisites.Hi-Fi.zip)

PR #988, opened 2019-05-22
gRPC improvement (#967)

Jakeeyturner:
Changes include:

- Updated extension 'failed to activate' to include error as well as a button to reload the window
- Updated `hasNativeDependenciesInstalled` to check if `grpc` can be required, to detect whether we need to rebuild
- Updated `npm rebuild` command to use the `--fallback-to-build` and `--target_arch` flag (set it to `process.arch` of the Code instance)


client/src/dependencies/DependencyManager.ts
```diff
@@ -39,7 +39,21 @@ export class DependencyManager {
...
+    // Need this function as proxyquire doesn't work
+    public requireGrpc(): void {
+        require('grpc');
+    }
+
     public async hasNativeDependenciesInstalled(): Promise<boolean> {
+        const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+
+        try {
+            this.requireGrpc();
```
cazfletch: This should loop through all the native dependencies in case we add more in the future

Jakeeyturner: Made it a bit more generic now!

PR #1030, opened 2019-05-31
Empty array for generated tests (close #1002)

### LINKED ISSUE #1002: Functional tests not generated correctly for params with array types
When the metadata for a parameter specifies the type as 'array' functional tests are not generated.

## Expected Behavior
It should generated a functional test as it does for types of string, boolean and number.

## Actual Behavior
Functional test not generated

## Possible Fix
Add an extra else if to handle arrays types, will need to recurse as sub type could also be an array as well as any other type or $ref.

## Steps to Reproduce
1. Create a function in chaincode with the `@Param` annotation specifying an array. For example `@Param('my_param', 'Array<String>')

Jakeeyturner:
When generating functional tests for a contract, if we detect a parameter is an array type, we will create an empty array as the value.

### DIFF
--- a/client/templates/tsTestSmartContractTemplate.ejs
+++ b/client/templates/tsTestSmartContractTemplate.ejs
@@ -96,6 +96,9 @@ describe('<%=chaincodeLabel%>' , () => {
                     <%_ } else if (parameter.schema.type === 'boolean') { _%>
                         <%_  params.push(` ${parameter.name.replace(`"`,'')}.toString()`) _%>
         const <%=parameter.name.replace(`"`,'')%>: <%=parameter.schema.type.replace(`"`,'')%> = true;
+                    <%_ } else if (parameter.schema.type === 'array') { _%>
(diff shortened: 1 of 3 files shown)


client/templates/tsTestSmartContractTemplate.ejs
```diff
@@ -96,6 +96,9 @@ describe('<%=chaincodeLabel%>' , () => {
                     <%_ } else if (parameter.schema.type === 'boolean') { _%>
                         <%_  params.push(` ${parameter.name.replace(`"`,'')}.toString()`) _%>
         const <%=parameter.name.replace(`"`,'')%>: <%=parameter.schema.type.replace(`"`,'')%> = true;
+                    <%_ } else if (parameter.schema.type === 'array') { _%>
+                        <%_  params.push(` JSON.stringify(${parameter.name.replace(`"`,'')})`) _%>
+        const <%=parameter.name.replace(`"`,'')%>: Array<any> = [];
```
cazfletch: This gives a tslint error as it wants any[]

Issue #928, opened 2019-05-07
Toggle Dev Mode - Display error


rthatcher:
Testing latest **v1.0.0** extension version, on **Ubuntu 18.04**, VS Code version **1.32.3**

I have a New Project (and no package or installed/instantiated code).
When I toggle dev mode I get this error message, but all seems happy (apart from the message)
<img width="530" alt="image" src="https://user-images.githubusercontent.com/29195399/57307976-65d43480-70dd-11e9-80d8-e2534c949a75.png">

heatherlp:
Hi @rthatcher can you reliably recreate this?

rthatcher:
Using the Market Place version (1.0.0) of the extension I now see different Notifications when toggling Development Mode.
![image](https://user-images.githubusercontent.com/29195399/58111736-f173b800-7be9-11e9-9a70-590665dba3c8.png)

I'm running with a VM rather than natively, so I see slightly slower performance than native users, but I have a relatively new machine with an i7 processor and an SSD drive.

I have seen these messages each time I have Toggled Development Mode today.

cazfletch:
this will be solved by #877

Jakeeyturner:
Issue fixed as a result of removing development mode in PR #1043

PR #1043, opened 2019-06-07
Removed toggling dev mode

### LINKED ISSUE #877: Tearing down fabric runtime in dev mode does not reset the dev mode icon
## Description
Tearing down fabric runtime does not reset the dev mode status of the peer/gateway, so when the fabric runtime is restarted, the UI reports being in the wrong mode 

## Expected Behavior

There are 2 options:
1. The fabric runtime is always started in 'normal' mode and the UI reflects that
2. The fabric runtime starts in the same mode as the last runtime to be torn down

## Your Environment

* Visual Studio Code version: 1.33.1
* Environment name and version Node.js v8.15.0, npm v6.4.1
* Operating System and version: MacOS 10.14.4

Jakeeyturner:
Close #877 and #928 


heatherlp:
I can't remember where this is at, but if we're keeping dev mode - can we update the peer tooltip to explain it's in dev mode when the infinity symbol is there? Helps explain what the infinity symbol actually is :)

client/src/fabric/FabricConnection.ts
```diff
@@ -197,4 +203,8 @@ export abstract class FabricConnection {
     private getAllPeers(): Array<Client.Peer> {
         return this.gateway.getClient().getPeersForOrg();
     }
+
+    private getClientTimeout(): number {
```
cazfletch: This shouldn't be done in here, it should be in the connect command and you pass the timeout through

client/src/debug/FabricDebugConfigurationProvider.ts
```diff
@@ -50,26 +51,26 @@ export abstract class FabricDebugConfigurationProvider implements vscode.DebugCo
                 return;
             }
 
-            if (!this.runtime.isDevelopmentMode()) {
+            const developmentMode: boolean = await this.runtime.isDevelopmentMode();
```
cazfletch: i think we can take this out. We need to update the generator version anyway so if we update the version here then they will have to tear down anyway and then always be in dev mode

client/src/fabric/FabricRuntime.ts
```diff
@@ -321,13 +321,21 @@ export class FabricRuntime extends EventEmitter {
         await this.killChaincodeInner(args, outputAdapter);
     }
 
-    public isDevelopmentMode(): boolean {
-        return this.developmentMode;
-    }
+    public async isDevelopmentMode(): Promise<boolean> {
```
cazfletch: see above comment about not needing this

client/src/util/MetadataUtil.ts
```diff
@@ -75,6 +75,7 @@ export class MetadataUtil {
 
             if (checkForEmpty && (contractsMap.size === 0)) {
                 outputAdapter.log(LogType.ERROR, `No metadata returned. Please ensure this smart contract is developed using the programming model delivered in Hyperledger Fabric v1.4+ for JavaScript and TypeScript`);
+                return null;
```
cazfletch: you don't need the null bit

client/README.md
```diff
@@ -122,6 +122,19 @@ The key `ecert` must be set to true in order for a smart contract to be able to
...
 Hovering over an identity in the `Fabric Wallets` panel will show any attributes associated with the identity.
 
+### Changing transaction timeout values
+
+In the extension, the default timeout value for transactions is 120 seconds.
+However, when debugging a transaction it might be necessary to change the time taken before a transaction fails.
+
+The following settings can be changed in the user settings to increase or decrease the time taken for a transaction to timeout:
+- `ibm-blockchain-platform.fabric.client.timeout` - timeout value used for communication to the peers and orderers
+- `ibm-blockchain-platform.fabric.chaincode.timeout` - timeout value for chaincode that's been started by the Local Fabric
```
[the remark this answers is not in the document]
Jakeeyturner: Updated the README to mention the two different variables we change

client/src/fabric/FabricConnection.ts
Jakeeyturner: Removed this function and now pass in the timeout value in the connect command (when calling connect)

client/src/debug/FabricDebugConfigurationProvider.ts
Jakeeyturner: Removed all signs of development mode!

PR #1054, opened 2019-06-13
Check generator version (close #1013)

### LINKED ISSUE #1013: Version fabric runtime
For example, if we upgrade the generator to Fabric v1.4.2, how do we get the user to go through the teardown/generate cycle? Do we ask them after they upgrade the extension? "There are new Fabric runtime features available. Do you want to upgrade? Yes/No/Not now"
and what happens we upgrade the generator in an incompatible way, so they have to go through the teardown/generate cycle

Jakeeyturner:

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -93,6 +93,9 @@ import { SettingConfigurations } from '../SettingConfigurations';
 import { exportWallet } from './commands/exportWalletCommand';
 import { UserInputUtil } from './commands/UserInputUtil';
 import { FabricWalletUtil } from './fabric/FabricWalletUtil';
+import { FabricRuntime } from './fabric/FabricRuntime';
+import { dependencies } from '../package.json';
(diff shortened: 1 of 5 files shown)


client/src/extension.ts
```diff
@@ -188,6 +194,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         outputAdapter.log(LogType.INFO, undefined, 'Tidying wallet settings');
         await FabricWalletUtil.tidyWalletSettings();
 
+        // Check if there is a newer version of the generator available
+        // This needs to be done as a seperate call to make sure the dependencies have been installed
+        const extensionData: ExtensionData = context.globalState.get<ExtensionData>(EXTENSION_DATA_KEY, DEFAULT_EXTENSION_DATA);
```
cazfletch: i don't think you need to get the data again

PR #1064, opened 2019-06-18
Debug version check

Jakeeyturner:


client/src/debug/FabricDebugConfigurationProvider.ts
```diff
@@ -50,24 +51,22 @@ export abstract class FabricDebugConfigurationProvider implements vscode.DebugCo
                 return;
             }
 
-            if (!this.runtime.isDevelopmentMode()) {
+            const developmentMode: boolean = this.runtime.isDevelopmentMode();
```
cazfletch: this shouldn't be in here, dev mode hasn't been updated in this version of the generator

PR #1166, opened 2019-07-22
Select CA for enrollment (Closes #1160)

### LINKED ISSUE #1160: Uses wrong CA for enrolling users via wallet interface
When creating a new identity using a gateway and enrollmentID and secret the extension selects the first certificate authority listed in the connection profile rather than the certificate authroity of the organization the connection profile refers too. This means that the enrollment fails as the extension attempts to register with another CA.

Steps to reproduce:
1. Have connection profile with multiple CA's with the first being not from the org listed at the top of the CP.
2. Create gateway using that connection profile
3. Create new wallet and add identity
4. Go through flow and get authorization failure as chosen CA knows nothing about the user you wish to register

toggle007:

### DIFF
--- a/client/src/commands/UserInputUtil.ts
+++ b/client/src/commands/UserInputUtil.ts
@@ -190,6 +190,18 @@ export class UserInputUtil {
         return vscode.window.showQuickPick(options, quickPickOptions);
     }
 
+    public static showQuickPickCA(listOfCAs: string[]): Thenable<string | undefined> {
+        const prompt: string = 'Choose your desired CA from the list' ;
+
+        const quickPickOptions: vscode.QuickPickOptions = {
+            ignoreFocusOut: true,
+            canPickMany: false,
+            placeHolder: prompt
+        };
+
+        return vscode.window.showQuickPick(listOfCAs, quickPickOptions);
+    }
+
     public static async showPeersQuickPickBox(prompt: string): Promise<string[] | undefined> {
         const connection: IFabricRuntimeConnection = await FabricRuntimeManager.instance().getConnection();
         const peerNames: Array<string> = connection.getAllPeerNames();
--- a/client/src/commands/addWalletIdentityCommand.ts
+++ b/client/src/commands/addWalletIdentityCommand.ts
@@ -12,6 +12,7 @@
  * limitations under the License.
 */
 'use strict';
+import * as yaml from 'js-yaml';
 import * as vscode from 'vscode';
 import * as fs from 'fs-extra';
 import { Reporter } from '../util/Reporter';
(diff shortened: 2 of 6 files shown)


client/src/commands/UserInputUtil.ts
```diff
@@ -66,7 +66,7 @@ export class UserInputUtil {
     static readonly IMPORT_WALLET: string = 'Specify an existing file system wallet';
     static readonly WALLET_NEW_ID: string = 'Create a new wallet and add an identity';
     static readonly ADD_IDENTITY_METHOD: string = 'Choose a method for adding an identity';
-    static readonly ADD_CERT_KEY_OPTION: string = 'Provide certificate and private key files';
+    static readonly  ADD_CERT_KEY_OPTION: string = 'Provide certificate and private key files';
```
Jakeeyturner: Could you remove the space you've added here please!

client/src/commands/addWalletIdentityCommand.ts
```diff
@@ -197,7 +193,33 @@ export async function addWalletIdentity(walletItem: WalletTreeItem | IFabricWall
...
+
+            if (gatewayRegistryEntry.connectionProfilePath.endsWith('.json')) {
+                connectionProfile = JSON.parse(connectionProfileFile);
+            } else {
+                // Assume its a yml/yaml file type
+                connectionProfile = yaml.safeLoad(connectionProfileFile);
+            }
+
+            // Get a list of CAs
+            const caKeys: any[] = Object.keys(connectionProfile.certificateAuthorities);
```
Jakeeyturner: Could you make the type `string[]` please

client/test/fabric/FabricCertificateAuthority.test.ts
```diff
@@ -43,46 +40,13 @@ describe('FabricCertificateAuthority', () => {
     describe('enroll', () => {
 
         it('should return certificate and key from enrollment using JSON connection profile', async () => {
```
Jakeeyturner: Should probably change the name of this test to be more accurate now that the code has changed.
E.g. `should return certificate and key from enrollment`

Jakeeyturner:
Could you also add a test for `addWalletIdentityCommand`, when the file read is a `yml` file

client/src/commands/addWalletIdentityCommand.ts
toggle007: Done

client/test/fabric/FabricCertificateAuthority.test.ts
toggle007: Done

PR #1173, opened 2019-07-23
fixedEvaluateStatement

toggle007:


client/cucumber/index.ts
```diff
@@ -76,7 +76,7 @@ async function runCucumberTest(): Promise<any> {
             features[4] = feature;
         } else if (file === 'transaction.feature') {
```
cazfletch: could we change transaction to submit please

toggle007: Changed it!

Issue #1105, opened 2019-07-01
Update display names of local gateway and wallet


cazfletch:
Currently the generator names the local fabric wallet as local_fabric_wallet and gateway as local_fabric. These should be changed in vscode to be Local Fabric and Local Fabric Wallet

PR #1176, opened 2019-07-24
UpdatedDisplayNames (Closes Issue #1105)

toggle007:


client/src/explorer/gatewayExplorer.ts
```diff
@@ -177,9 +178,12 @@ export class BlockchainGatewayExplorerProvider implements BlockchainExplorerProv
                     arguments: [runtimeGateway]
                 };
 
+                let gatewayName: string = runtimeGateway.name;
```
cazfletch: you don't need both these lines you only need to set the gateway name

toggle007: Done

Issue #1179, opened 2019-07-24
Suggestion of menu working change


mbwhite:
## Description
From drop down menu from the 'Smart Contract Package' view it's easy to pick the wrong option even when you know exactly what you want. 

Changing the wording could help here. It's complicated in a small way as the concept of 'packages' and 'projects' are not VSCode conccepts

## Possible Fix
`Create new Smart Contract project`
`Package Smart Contract project` OR `Create a new Smart Contract package`
`Import Smart Contract package`

## Steps to Reproduce

1.
2.
3.

## Existing issues

- [ ] [Stack Overflow issues](https://stackoverflow.com/questions/tagged/ibp-vscode-extension)
- [ ] [GitHub Issues](https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues)

## Your Environment

* IBM Blockchain Platform Extension Version:
* Visual Studio Code version:
* Environment name and version (e.g. Node.js v8.12.0, npm v6.4.1):
* Operating System and version:
* Link to your project:

toggle007:
Merged

PR #1183, opened 2019-07-25
Changed Menu Labels (Closes Issue #1179)

toggle007:


toggle007:
Updated labels as discussed.

toggle007:
> The labels are now correct but the items are in the wrong order
The item order couldn't be changed as these were dependant on the LabelName and not the command name as discussed.

PR #1204, opened 2019-07-31
Discover electron version

### LINKED ISSUE #1134: Remove hardcoded Electron/Node versions where possible
Today, the extension hard codes versions of native modules into the package.json:

```
    "nativeDependencies": {
        "grpc": {
            "linux-ia32": {
                "original": "./node_modules/grpc/src/node/extension_binary/node-v64-linux-ia32-glibc",
                "new": "./node_modules/grpc/src/node/extension_binary/electron-v3.0-linux-ia32-glibc"
            },
            "linux-x64": {
                "original": "./node_modules/grpc/src/node/extension_binary/node-v64-linux-x64-glibc",
                "new": "./node_modules/grpc/src/node/extension_binary/electron-v3.0-linux-x64-glibc"
            },
            "darwin-x64": {
                "original": "./node_modules/grpc/src/node/extension_binary/node-v64-darwin-x64-unknown",
                "new": "./node_modules/grpc/src/node/extension_binary/electron-v3.0-darwin-x64-unknown"
            },
            "win32-ia32": {
                "original": "./node_modules/grpc/src/node/extension_binary/node-v64-win32-ia32-unknown",
                "new": "./node_modules/grpc/src/node/extension_binary/electron-v3.0-win32-ia32-unknown"
            },
            "win32-x64": {
                "original": "./node_modules/grpc/src/node/extension_binary/node-v64-win32-x64-unknown",
                "new": "./node_modules/grpc/src/node/extension_binary/electron-v3.0-win32-x64-unknown"
            }
        }
    },
```

The versions are:
- The Electron version, `v3.0`
- The Node.js module version, `v64`

This is fragile and means that whenever VSCode picks up a new version of Electron/Node.js, the extension breaks, and we have to publish a new version. 

We should be able to figure out all this information programatically at runtime instead.

The Node.js module version can be determined by inspecting `process.versions`:

```
$ node -pe process.versions
{ http_parser: '2.8.0',
  node: '10.16.0',
  v8: '6.8.275.32-node.52',
  uv: '1.28.0',
  zlib: '1.2.11',
  brotli: '1.0.7',
  ares: '1.15.0',
  modules: '64', <-- this one
  nghttp2: '1.34.0',
  napi: '4',
  openssl: '1.1.1b',
  icu: '64.2',
  unicode: '12.1',
  cldr: '35.1',
  tz: '2019a' }
```

The Electron version should also be in this list, but it is not 😢 

There is a VSCode issue open for this: https://github.com/microsoft/vscode/issues/62315

An alternative (as opposed to fixing that issue ourselves, or waiting for it to be fixed) is attempting to figure out what Electron version we're using from the information we do have available. Electron publish a list of Electron versions and all of their built-in dependencies - information we do have from `process.versions` - see this file:

https://github.com/electron/releases/blob/master/lite.json

cazfletch:
Discover the version of electron so we automatically update grpc and don't have to hard code paths

closes #1134


client/src/dependencies/DependencyManager.ts
```diff
@@ -130,25 +124,72 @@ export class DependencyManager {
...
+                await fs.rename(newPath, origPath);
             }
 
             outputAdapter.log(LogType.SUCCESS, undefined, 'Finished updating native node modules');
             progress.report({message: 'Finished updating native node modules'});
         });
     }
 
+    private async getRebuildInfo(os: string, arch: string, thing: string): Promise<{ modules: string, longVersion: string, shortVersion: string }> {
+        const response: any = await Axios.get('https://raw.githubusercontent.com/electron/releases/master/lite.json');
```
Jakeeyturner: Should we be try/catching this, or even setting a timeout for how long this should take? What happens if the user doesn't have an internet connection?

cazfletch: If the user doesn't have an internet connection then none of the getting the grpc prebuilt binary code will work. The error will eventually be caught by failing to activate.

PR #1213, opened 2019-08-02
Debug tutorial

cazfletch:


Charlotte8699:
I've worked through the tutorial and these are the notes I made while doing it.

Stage 1: Next to the code that you have to copy and paste, have a copy button
Would make it easier for the user rather than highlighting and copying it.

Stage 2: Fabric runtime needs to be started - this isn't stated in the tutorial
I received an error message indicating the fabric runtime needed to be started before proceeding.

Other than that, it was easy enough to work through! I liked how concise and to-the-point it was.

Issue #1265, opened 2019-08-08
Extension still shows local_fabric at some places instead of Local Fabric


toggle007:
## Description
Should say 'Connecting to Local Fabric' and not 'Connecting to local_fabric'
Same with connected to local_fabric... when connecting with a given identity.

## Steps to Reproduce

1.
2.
3.

## Existing issues

- [ ] [Stack Overflow issues](https://stackoverflow.com/questions/tagged/ibp-vscode-extension)
- [ ] [GitHub Issues](https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues)

## Your Environment

* IBM Blockchain Platform Extension Version:
* Visual Studio Code version:
* Environment name and version (e.g. Node.js v8.12.0, npm v6.4.1):
* Operating System and version:
* Link to your project:

PR #1268, opened 2019-08-08
Updated log messages to say Local fabric and not local_fabric (#1265)

toggle007:
Also added relevant tests.


client/src/commands/environmentConnectCommand.ts
```diff
@@ -85,7 +86,14 @@ export async function fabricEnvironmentConnect(fabricEnvironmentRegistryEntry: F
 
         FabricEnvironmentManager.instance().connect(connection, fabricEnvironmentRegistryEntry);
 
-        outputAdapter.log(LogType.SUCCESS, `Connected to ${fabricEnvironmentRegistryEntry.name}`, `Connected to ${fabricEnvironmentRegistryEntry.name}`);
+        let gatewayName: string;
```
cazfletch: this isn't a gateway you need to update the variable name

PR #1282, opened 2019-08-12
Throw error if no peers

### LINKED ISSUE #1098: Gateway with invalid MSP does not show any errors
If you import a gateway with the following `organizations` section:

```
    "organizations": {
        "DemoOrg1": {
            "mspid": "DemoOrg1",
            "peers": [
                "n5162fb-demoorg1peer2.sjc04cluster.sjc04.containers.appdomain.cloud:7051"
            ],
            "certificateAuthorities": [
                "n5162fb-demoorg1ca.sjc04cluster.sjc04.containers.appdomain.cloud:7054"
            ]
        }
    },
```

And then you add/import an identity with a different MSP:

```
{"name":"demorg1admin","mspid":"DemoOrg1MSP","roles":null,"affiliation":"", ...
```

And then you connect to that gateway with that identity, the connection succeeds, but nothing appears under channels.

This is because the MSP IDs don't match, so the extension doesn't find any peers for that gateway, and then doesn't look for any channels.

If we don't find any peers to query the list of channels from, we should throw an error to tell the user that something is wrong.

cazfletch:
Throw an error if can't find any peers to query channels from

closes #1098

### DIFF
--- a/client/test/fabric/FabricConnection.test.ts
+++ b/client/test/fabric/FabricConnection.test.ts
@@ -104,7 +104,7 @@ describe('FabricConnection', () => {
 
         fabricClientStub.getMspid.returns('myMSPId');
         fabricCAStub = mySandBox.createStubInstance(fabricClientCA);
-        fabricCAStub.enroll.returns({certificate : 'myCert', key : { toBytes : mySandBox.stub().returns('myKey')}});
+        fabricCAStub.enroll.returns({ certificate: 'myCert', key: { toBytes: mySandBox.stub().returns('myKey') } });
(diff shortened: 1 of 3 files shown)


client/test/fabric/FabricConnection.test.ts
```diff
@@ -591,11 +591,17 @@ describe('FabricConnection', () => {
             await fabricConnection.createChannelMap().should.be.rejectedWith(`Cannot connect to Fabric: ${error.message}`);
         });
 
+        it('should handle no peeros', async () => {
```
Jakeeyturner: `peers`

PR #1296, opened 2019-08-14
Don't show multiple ordering nodes

### LINKED ISSUE #1229: Ordering node clusters split into multiple individual nodes in the tree
When you import an ordering node from the IBPv2 SaaS/software, and that ordering node is a cluster of 5 ordering nodes, you end up with 5 ordering nodes in a tree.

Ordering node clusters should be presented as a single node in the tree, and when we interact with it, one (any will do) of the nodes in the cluster should be used.

There is a `cluster_name` property in the node JSON file that we should use instead of the `name` property.

cazfletch:
If cluster_name is set on an ordering node then all the ordering nodes in that cluster will be grouped together

closes #1229

### DIFF
--- a/client/test/fabric/FabricEnvironment.test.ts
+++ b/client/test/fabric/FabricEnvironment.test.ts
@@ -169,6 +169,41 @@ describe('FabricEnvironment', () => {
 
             writeStub.should.have.been.calledWith(nodePath, node);
         });
+
+        it('should update orderer node with cluster name', async () => {
+            const node: FabricNode = FabricNode.newOrderer('order', 'orderer.example.com', 'http://localhost:17056', undefined, undefined, 'osmsp', 'myCluster');
+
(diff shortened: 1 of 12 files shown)


client/test/fabric/FabricEnvironment.test.ts
```diff
@@ -169,6 +169,41 @@ describe('FabricEnvironment', () => {
...
+
+            const nodePath: string = path.join(environmentPath, 'nodes', `${node.name}.json`);
+            const writeStub: sinon.SinonStub = sandbox.stub(fs, 'writeJson');
+
+            await environment.updateNode(node);
+
+            writeStub.should.have.been.calledWith(nodePath, node);
+        });
+
+        it('should update all orderer nodes in same cluster', async () => {
```
Jakeeyturner: Can we separate out a test for 'handling multiple orderer services'?

PR #1326, opened 2019-08-23
Added validation for asset type

toggle007:


src/commands/createSmartContractProjectCommand.ts
```diff
@@ -55,10 +55,16 @@ export async function createSmartContractProject(): Promise<void> {
...
     if (smartContractLanguageItem.type === LanguageType.CONTRACT) {
         assetType = await UserInputUtil.showInputBox('Name the type of asset managed by this smart contract', 'MyAsset');
+        const regexForAssetType: RegExp = /^[A-Z]+$/i;
+        const validAssetType: boolean = regexForAssetType.test(assetType);
         if (!assetType) {
             // User has cancelled the input box
             return;
         }
+        if (!validAssetType) {
+            outputAdapter.log(LogType.ERROR, `Please ensure your asset name only consists of alphabets.`);
```
cazfletch: I think this error message needs changing - maybe to something like `Invalid asset name, it should only contain lowercase and uppercase letters`

toggle007: Made the changes, thanks for the feedback!

PR #1415, opened 2019-09-17
Added description to package smart contract command

### LINKED ISSUE #1399: Provide greyed out helpers (dir paths) for workspace folders in Package command
I accidentally added two "contract" folders from commercial paper (different locations, one for DigiBank, one for MagnetoCorp) into my workspace. When I came to run the Package Command I got a screenshot that asked me whether I wanted to package "contract" or "contract". I guessed that the list was in the same order as the folders in my workspace. 

Could the Package command be updated to have some greyed out helpers after the folders to show their full paths? 

Jake knows what I mean. Might be relevant for other commands too?

image

toggle007:
Path of the contract now in grey so can distinguish the contracts (if they're named the same but have different paths).
#1399 

### DIFF
--- a/test/commands/UserInputUtil.test.ts
+++ b/test/commands/UserInputUtil.test.ts
@@ -857,18 +857,20 @@ describe('UserInputUtil', () => {
     });
 
     describe('showWorkspaceQuickPickBox', () => {
-        it('should show the workspace folders', async () => {
+        it('should show the workspace folders and their path as a description', async () => {
             mySandBox.stub(UserInputUtil, 'getWorkspaceFolders').returns([
                 {
                     name: 'myPath1',
                     uri: {
-                        path: 'path1'
+                        path: 'path1',
+                        fsPath: 'pathHere'
                     }
                 },
                 {
                     name: 'myPath2',
                     uri: {
-                        path: 'path2'
+                        path: 'path2',
+                        fsPath: 'pathHere'
                     }
                 }]);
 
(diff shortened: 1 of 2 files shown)


test/commands/UserInputUtil.test.ts
```diff
@@ -898,6 +898,48 @@ describe('UserInputUtil', () => {
                 placeHolder: 'choose a folder'
             });
         });
+
+        it('should show the workspace folders along with their path as a description', async () => {
```
cazfletch: you don't need a separate test you can just add to the existing one as they both do the same thing

test/commands/UserInputUtil.test.ts
```diff
@@ -898,6 +898,48 @@ describe('UserInputUtil', () => {
...
+            result.should.deep.equal({
+                label: 'myPath1', data: {
+                    name: 'myPath1',
+                    uri: {
+                        path: 'path1'
+                    }
+                }, description : 'pathHere'
+            });
+
+            quickPickStub.should.have.been.calledWith(sinon.match.any, {
```
cazfletch: you should change `sinon.match.any` to be the quick pick items that should be shown so you can test the it shows the correct description

test/commands/UserInputUtil.test.ts
toggle007: Done

test/commands/UserInputUtil.test.ts
toggle007: I believe I have now got the correct quick pick items - test passes.

PR #1430, opened 2019-09-23
Can export a connection profile while connected to a gateway

### LINKED ISSUE #1386: Gateway must be disconnected to export connection profile
In order to have a right-click option to export a connection profile, the gateway must be in a disconnected state.

## Description
When trying to export a connection profile, the option was not available to do so. The option only appeared when the gateway was disconnected.

## Expected Behavior
Export option should be available irrespective of gateway connection state.

## Actual Behavior
No option available

## Steps to Reproduce
Connect gateway, try to export connection profile.

### LINKED ISSUE #1444: Local Fabric gateway title appears differently in the gateways view vs the quick pick menu
## Description
There's inconsistency in how we display the title of the Local Fabric gateway in the gateways view vs the quick pick menu, which we should fix.

## Expected Behavior
Title consistently appears as "Local Fabric"

## Actual Behavior
Title appears as "local_fabric" in the quick pick menu

## Steps to Reproduce

1. Run the Export Connection Profile command from the command palette
2. See "local_fabric" instead of "Local Fabric"

## Existing issues

## Your Environment

* Visual Studio Code version:
* Link to your project:

erin-hughes:
I've updated the gateway panel so that when you are connected to a gateway, there is an option at the top of the panel that a user can click on to export their connection profile. This is consistent with the behaviour seen in the environments panel.

Closes #1386. Also closes #1444  

Signed-off-by: Erin <Erin.Hughes@ibm.com>


src/commands/exportConnectionProfileConnectedCommand.ts
```diff
@@ -0,0 +1,27 @@
+/*
```
cazfletch: don't need this class as can do the samething from where you register the command

Issue #1443, opened 2019-09-25
Submit transaction not working correctly when same contract on two channels


cazfletch:
If you have 2 channels with a smart contract deployed to both with the same name then submit transaction doesn't always go to the right channel

cazfletch:
Testing: Create a network with 2 channels and two peers. Install and instantiate the same chaincode on two peers and two channels. Check it only appears one in the installed chaincode list, check that it only appears once in the instantiated chaincode list

PR #1458, opened 2019-09-27
Fix duplicate instantiated contracts

cazfletch:
Updated so that if you have the same contract deployed on two channels it will only show once in the tree
Updated tooltip to show which channels it's instantiated on

contributes to #1443

### DIFF
--- a/extension/explorer/model/InstantiatedTreeItem.ts
+++ b/extension/explorer/model/InstantiatedTreeItem.ts
@@ -27,10 +27,17 @@ export abstract class InstantiatedTreeItem extends BlockchainTreeItem {
         dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', 'smart-contract.svg')
     };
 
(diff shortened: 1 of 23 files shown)


extension/explorer/model/InstantiatedTreeItem.ts
```diff
@@ -27,10 +27,17 @@ export abstract class InstantiatedTreeItem extends BlockchainTreeItem {
         dark: path.join(__filename, '..', '..', '..', '..', '..', 'resources', 'dark', 'smart-contract.svg')
     };
 
-    constructor(provider: BlockchainExplorerProvider, public readonly name: string, public readonly channel: ChannelTreeItem, public readonly version: string, public readonly collapsibleState: vscode.TreeItemCollapsibleState, public readonly contracts?: string[], public readonly showIcon?: boolean) {
+    constructor(provider: BlockchainExplorerProvider, public readonly name: string, public channels: ChannelTreeItem[], public readonly version: string, public readonly collapsibleState: vscode.TreeItemCollapsibleState, public readonly contracts?: string[], public readonly showIcon?: boolean) {
```
lquintai: should channels still be readonly ?

PR #1470, opened 2019-10-01
Fix duplicate installed contracts

cazfletch:
Updated so that if you have multiple peers with that same contract installed it will only show once
Updated to tooltip to show which peers a contract is installed on
Fixed linting

contributes to #1443


test/explorer/environmentExplorer.test.ts
```diff
@@ -736,26 +738,27 @@ describe('environmentExplorer', () => {
                 const contractTreeItems: Array<BlockchainTreeItem> = await blockchainRuntimeExplorerProvider.getChildren(allChildren[1]);
                 const installedContractsTree: Array<BlockchainTreeItem> = await blockchainRuntimeExplorerProvider.getChildren(contractTreeItems[0]);
                 installedContractsTree.length.should.equal(5);
-                const installedContractOne: InstantiateCommandTreeItem = installedContractsTree[0] as InstantiateCommandTreeItem;
+                const installedContractOne: InstantiateCommandTreeItem = installedContractsTree[0] as InstalledChainCodeOpsTreeItem;
```
lquintai: This and other instances of InstantiateCommandTreeItem should be InstalledChainCodeOpsTreeItem instead. Only leaving one comment for all of these.

PR #1505, opened 2019-10-08
Only show debug icon on contract debug (close #1464)

### LINKED ISSUE #1464: Blockchain debug toolbar icon appearing when it shouldn't
The blockchain icon is appearing on the debug toolbar even though not debugging a smart contract we should check that we are definitely debugging a smart contract

Jakeeyturner:

### DIFF
--- a/extension/debug/FabricDebugConfigurationProvider.ts
+++ b/extension/debug/FabricDebugConfigurationProvider.ts
@@ -29,6 +29,8 @@ import { FabricChaincode } from '../fabric/FabricChaincode';
 
 export abstract class FabricDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
 
+    static readonly debugEvent: string = 'contractDebugging';
+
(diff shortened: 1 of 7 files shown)


extension/debug/FabricDebugConfigurationProvider.ts
```diff
@@ -138,6 +138,10 @@ export abstract class FabricDebugConfigurationProvider implements vscode.DebugCo
             // configuration provider, for example a fabric:go configuration with the go debug configuration
             // provider. This results in errors and we need to just force it to use our configuration as-is.
             delete resolvedConfig.name;
+
+            // We need this in order to differentiate between debug events
+            resolvedConfig.debugEvent = 'contractDebugging';
```
cazfletch: Can we make this a static variable please

Jakeeyturner: Done!

PR #1530, opened 2019-10-14
Transaction peer targeting (close #1514)

### LINKED ISSUE #1514: As a user i can submit transactions and use peer targetting
Issue to cover being able to target which peers to target when submitting a transaction

Jakeeyturner:

### DIFF
--- a/extension/commands/submitTransaction.ts
+++ b/extension/commands/submitTransaction.ts
@@ -24,6 +24,7 @@ import { VSCodeBlockchainDockerOutputAdapter } from '../logging/VSCodeBlockchain
 import { InstantiatedTreeItem } from '../explorer/model/InstantiatedTreeItem';
 import { FabricGatewayRegistryEntry } from '../registries/FabricGatewayRegistryEntry';
 import { FabricRuntimeUtil } from '../fabric/FabricRuntimeUtil';
+import { IFabricClientConnection } from '../fabric/IFabricClientConnection';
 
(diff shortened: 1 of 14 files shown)


extension/commands/submitTransaction.ts
```diff
@@ -120,6 +121,25 @@ export async function submitTransaction(evaluate: boolean, treeItem?: Instantiat
         return;
     }
 
+    const connection: IFabricClientConnection = FabricConnectionManager.instance().getConnection();
+    const channelPeerNames: string[] = await connection.getChannelPeerNames(channelName);
```
cazfletch: why are you getting the peer names before you know if you even need to

extension/fabric/FabricConnection.ts
```diff
@@ -136,6 +136,44 @@ export abstract class FabricConnection {
...
+
+            return channelPeers;
+        } catch (error) {
+            throw new Error(`Unable to get channel peers: ${error.message}`);
+        }
+    }
+
+    public async getChannelPeerNames(channelName: string): Promise<string[]> {
+        try {
+            const network: Network = await this.gateway.getNetwork(channelName);
```
cazfletch: can this function call the function above rather than repeating code?

extension/fabric/FabricConnection.ts
```diff
@@ -136,6 +136,44 @@ export abstract class FabricConnection {
         }
     }
 
+    public async getChannelPeers(channelName: string, peerNames: string[]): Promise<Client.ChannelPeer[]> {
```
cazfletch: this function doesn't need to be public you only use it in the child class you can just make it protected

PR #1552, opened 2019-10-15
Associate multiple nodes

### LINKED ISSUE #1416: Associate identity with more nodes should use check box
When you want to associate an identity with more than one node this should be changed to ask if you want to and then show you a list of nodes with check boxes of nodes that you can check the ones you want to associate

toggle007:
Added functionality and tests to associate multiple nodes at once using checkbox
Closes #1416 

### DIFF
--- a/extension/commands/associateIdentityWithNode.ts
+++ b/extension/commands/associateIdentityWithNode.ts
@@ -158,27 +158,27 @@ export async function associateIdentityWithNode(replace: boolean = false, enviro
         vscode.commands.executeCommand(ExtensionCommands.CONNECT_TO_ENVIRONMENT, environmentRegistryEntry);
         outputAdapter.log(LogType.SUCCESS, `Successfully associated identity ${node.identity} from wallet ${node.wallet} with node ${node.name}`);
 
-        let askAgain: boolean = true;
         let otherNodes: FabricNode[] = [];
-        do {
-            otherNodes = await environment.getNodes(true);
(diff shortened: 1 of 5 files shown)


extension/commands/associateIdentityWithNode.ts
```diff
@@ -168,28 +168,31 @@ export async function associateIdentityWithNode(replace: boolean = false, enviro
...
 
-            items.push('No');
-            const nodeName: string = await UserInputUtil.showQuickPick('Do you want to associate the same identity with another node?', items) as string;
-
-            if (!nodeName || nodeName === 'No') {
+            if (!yesnoPick || yesnoPick === UserInputUtil.NO) {
                 askAgain = false;
             } else {
-                const foundNode: FabricNode = otherNodes.find((_node: FabricNode) => nodeName.endsWith(_node.name));
+                    const nodes: IBlockchainQuickPickItem<FabricNode>[] = await UserInputUtil.showFabricNodeQuickPick('Choose the nodes you wish to associate with this identity', environmentRegistryEntry.name, [], false, true) as IBlockchainQuickPickItem<FabricNode>[];
```
cazfletch: this doesn't work as it shows you all the nodes so you are shown all the ones that already have identities and the one that you just assigned an identity too. You need to use the nodes in `otherNodes`.

toggle007: Fixed

Issue #1517, opened 2019-10-10
Update where the registries are stored


cazfletch:
All the registries should be stored on disk rather than in the vscode settings as the settings file is shared across all environments

PR #1562, opened 2019-10-17
Update environment registries

cazfletch:
Update to make environment registries use the file based registry

contributes to #1517


extension/fabric/FabricRuntime.ts
```diff
@@ -292,7 +299,7 @@ export class FabricRuntime extends FabricEnvironment {
     }
 
     public async isCreated(): Promise<boolean> {
-        return await fs.pathExists(this.path);
+        return FabricEnvironmentRegistry.instance().exists(this.name);
```
Jakeeyturner: Do we need to `await` this?

cazfletch: no as the await on isCreated will wait

PR #1568, opened 2019-10-18
Added tree item for empty panels (close #1564)

### LINKED ISSUE #1564: Add copy to validly-empty UI panels
In #1498 we are adding the option to "not use Local Fabric". This would mean it's possible to arrive at the IBP panel with the extension installed and absolutely nothing to display in any of the 4 view panels on the left!

Feels like this will look like "it hasn't loaded properly" (well, in fact it's exactly what it looks like when it doesn't load properly haha). So, how about we add some text to the panels? Suggest that if a panel is empty we display its text (and then as soon as there's anything else to display, that replaces the text).

Suggestion of how this could look:
image

Jakeeyturner:

### DIFF
--- a/extension/explorer/model/TextTreeItem.ts
+++ b/extension/explorer/model/TextTreeItem.ts
@@ -1,0 +1,26 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+'use strict';
+
+import * as vscode from 'vscode';
+import { BlockchainTreeItem } from './BlockchainTreeItem';
+import { BlockchainExplorerProvider } from '../BlockchainExplorerProvider';
+
+export class TextTreeItem extends BlockchainTreeItem {
+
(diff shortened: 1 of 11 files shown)


test/explorer/packageExplorer.test.ts
```diff
@@ -63,6 +64,14 @@ describe('packageExplorer', () => {
         testPackages[2].label.should.equal('vscode-pkg-3@1.2.3');
         logSpy.should.not.have.been.calledWith(LogType.ERROR);
     });
+
+    it('should say that there are no packages', async () => {
+        mySandBox.stub(PackageRegistry.instance(), 'getAll').resolves([]);
```
cazfletch: getAll shouldn't be stubbed you can just call clear

extension/explorer/model/TextTreeItem.ts
```diff
@@ -0,0 +1,26 @@
...
+'use strict';
+
+import * as vscode from 'vscode';
+import { BlockchainTreeItem } from './BlockchainTreeItem';
+import { BlockchainExplorerProvider } from '../BlockchainExplorerProvider';
+
+export class TextTreeItem extends BlockchainTreeItem {
+
+    constructor(provider: BlockchainExplorerProvider, public readonly label: string) {
+        super(provider, label, 0);
```
cazfletch: can you use `vscode.TreeItemCollapsibleState.None` here please

test/registries/PackageRegistry.test.ts
```diff
@@ -156,4 +156,11 @@ describe('PackageRegistry', () => {
 
     });
 
+    describe('#clear', () => {
+        it('should clear all packages', async () => {
+            const emptyDirStub: sinon.SinonStub = mySandBox.stub(fs, 'emptyDir').resolves();
```
cazfletch: you don't need to stub this. If you add a package you can then check there are none when you call getAll

PR #1576, opened 2019-10-22
Enable remote development to open local files

### LINKED ISSUE #1575: Enable local files to be opened when doing remote development
In order to make show local files work when doing remote development and doing an action with open Dialog we need to use the fs commands from vscode rather than fs-extra

To test this, you will need to have a machine you can ssh into this can be a vm. Then try doing actions that let you choose files and choose the option to show local files

cazfletch:
Updated to use vscode fs commands rather than fs extra when we are doing actions after choosing a file using the open dialog box

Note this won't work for creating smart contracts or importing wallets. An error is shown if you try and do this

Also updated README

closes #1575


README.md
```diff
@@ -53,7 +53,17 @@ You can check your installed versions by running the following commands from a t
 
 To open the Prerequisites page again, run the following in the command palette: `View Prerequisites`
 
-Please note: the extension doesn't currently work with the VSCode Remote Development feature, we plan to make this work in the future, follow progress [here](https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/1357). 
+## Remote Development
+### SSH
+The extension supports remote development using SSH. Any vscode settings are shared between local and remote. If you update the setting `ibm-blockchain-platform.ext.directory` then please be aware that you will get an error if the path isn't valid for the environment you are using the extension in.
```
Jakeeyturner: `VS Code` instead of `vscode`

packages/blockchain-extension/extension/commands/UserInputUtil.ts
```diff
@@ -740,7 +740,7 @@ export class UserInputUtil {
 
     }
 
-    public static async browse(placeHolder: string, quickPickItems: string[] | { label: string, description: string }[], openDialogOptions: vscode.OpenDialogOptions, returnUri?: boolean): Promise<string | vscode.Uri | vscode.Uri[]> {
+    public static async browse(placeHolder: string, quickPickItems: string[] | { label: string, description: string }[], openDialogOptions: vscode.OpenDialogOptions): Promise<vscode.Uri | vscode.Uri[]> {
```
Jakeeyturner: Just noticed that we have to pass `openDialogOptions` through every time we call `browse`.
It would be nice if the function had some default options and we just passed flags through instead.

packages/blockchain-extension/extension/commands/addWalletIdentityCommand.ts
```diff
@@ -121,22 +119,22 @@ export async function addWalletIdentity(walletItem: WalletTreeItem | FabricWalle
...
-            const jsonIdentityPath: vscode.Uri = await UserInputUtil.browse('Browse for a JSON identity file', [UserInputUtil.BROWSE_LABEL], openDialogOptions, true) as vscode.Uri;
+            const jsonIdentityPath: vscode.Uri = await UserInputUtil.browse('Browse for a JSON identity file', [UserInputUtil.BROWSE_LABEL], openDialogOptions) as vscode.Uri;
             if (!jsonIdentityPath) {
                 return;
             }
             const certProperty: string = 'cert';
             const privateKeyProperty: string = 'private_key';
-            const jsonIdentityContents: string = await fs.readFile(jsonIdentityPath.fsPath, 'utf8');
-            const jsonIdentity: any = JSON.parse(jsonIdentityContents);
+            const jsonIdentityContents: Uint8Array = await vscode.workspace.fs.readFile(jsonIdentityPath);
```
Jakeeyturner: Did we want to put a wrapper around this?
E.g.
```
static async ExtensionUtil.readFile(path: string): Promise<string> {
    const fileContent: Uint8Array = await vscode.workspace.fs.readFile(path);
    return fileContent.toString();
}
```

I think we're likely to remember calling ExtensionUtil.readFile() over vscode.workspace.fs.readFile()

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -90,7 +90,9 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
         let addedAllNodes: boolean = true;
         for (const nodeUri of nodeUris) {
             try {
-                let nodes: FabricNode | Array<FabricNode> = await fs.readJson(nodeUri.fsPath);
+                const nodeContents: Uint8Array = await vscode.workspace.fs.readFile(nodeUri);
+                const nodeString: string = nodeContents.toString();
+                let nodes: FabricNode | Array<FabricNode> = JSON.parse(nodeString);
```
Jakeeyturner: Following on from the previous comment we could even have ExtensionUtil.readJSON(path).

packages/blockchain-extension/extension/fabric/FabricCertificate.ts
```diff
@@ -43,8 +41,9 @@ export class FabricCertificate {
 
     }
 
-    public static loadFileFromDisk(path: string): string {
-        return fs.readFileSync(path, ENCODING) as string;
+    public static async loadFileFromDisk(uri: vscode.Uri): Promise<string> {
```
Jakeeyturner: If we do the following suggestions we could remove this function.

Issue #1155, opened 2019-07-17
improve submit transaction


cazfletch:
Submitting a transaction is ok if you just have strings as parameters but it is not ideal if you have objects especially if they are large objects. We should provide a way of submitting a transaction more easily with objects. It would also be good to be able to repeatedly submit transactions without having to go through the process of typing in the parameters again.

EdMoffatt:
Another one for you Charlotte. I think Andy Hurt or Liam Grace would be good folks to explain to you what's needed here :) Similar plan to the events one, do some research, come up with some suggestions, and then let's review both of them in one meeting with the VS Code squad :+1:

awjh-ibm:
Would be great if my previous inputs were selectable below (in the style of the selecting a function) and also if the form could tell me the arg names not just to enter arg0 etc

Charlotte8699:
Transaction web view update

This is the latest click through of the transactions:
https://xd.adobe.com/view/7b0c6b7e-21b5-46fa-64d7-14b5d4c17091-0d3d/

I've since had further feedback from Liam which I'll be acting on today/tomorrow. I have also been working with @erin-hughes to find out what i need to be doing to make creating this web view easier for her.

EdMoffatt:
Great progress on this @Charlotte8699 . I wonder how it would look without the twisties in the "Recent" view? I think a permanently displayed table could work here, with new transactions appearing at the top - collapsing by date doesn't seem to add much to me!

awjh-ibm:
Has any thought been put into how i might run this using keyboard shortcuts? Ideally I would like to look at my code whilst I run the command. I usually do most of my repeatable transactions when i am in debug mode and like to stay on the same code page so I can view the values in variables as it goes through breakpoints. 

The current method of runnign transactions is nice as I can do it entirely from my keyboard. CMD+L I have configured to bring up the Evaluate vs Submit options and can then use arrow keys and enter to select the transaction and run it.

cazfletch:
I don't think you would want the alternatives for arguments as you don't have key value pairs for arguments, this would work for transient data though. It would also need an `ecert` property (this can either be true or false) as well as the name and value properties you have

Jakeeyturner:
With the work happening in #1510, we'll also need to be able to handle peer-targeting in the new transaction page.

Charlotte8699:
[...]

PR #1608, opened 2019-11-05
Generation of transaction templates when a user selects a transaction

### LINKED ISSUE #1483: As a user I can create a new transaction template
The user should be able to choose which transaction that they want to create this template for via a dropdown.

image

The template should include the names of the parameters that a user needs to provide in order for the transaction to be successful.

erin-hughes:
Closes #1483 and contributes to #1155

### DIFF
--- a/enzyme/tests/App.test.tsx
+++ b/enzyme/tests/App.test.tsx
@@ -1,0 +1,122 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import App from '../../src/App';
+import chai from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+import ISmartContract from '../../src/interfaces/ISmartContract';
+import ITransaction from '../../src/interfaces/ITransaction';
+import Utils from '../../src/Utils';
+chai.should();
+chai.use(sinonChai);
+
+describe('App', () => {
+
+    let mySandBox: sinon.SinonSandbox;
+
+    const mockTxn: ITransaction = {
+        name: 'mockTxn',
+        parameters: [{
+            description: '',
+            name: 'name',
+            schema: {}
+        }],
+        returns: {
+            type: ''
+        },
+        tag: ['submit']
+    };
+
+    const greenContract: ISmartContract = {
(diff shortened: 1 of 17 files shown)


enzyme/tests/App.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+            }
+        });
+        dispatchEvent(msg);
+        expect(component.state().childState).toBe(mockState);
+
+        Utils.changeRoute('/transaction/create');
+        expect(component.state().childState).toBe(mockState);
+    });
+
+    it('updates the state correctly when switching smart contracts', async () => {
```
Jakeeyturner: Not sure this belongs in `App.test.tsx` as it's for testing specifically of the transaction home view.

enzyme/tests/TransactionCreate.test.tsx
```diff
@@ -1,18 +1,61 @@
-
+// tslint:disable no-unused-expression
```
Jakeeyturner: Would it be possible to have a look at how we get our tslint rules working with the React stuff - that way we wouldn't need to do inline stuff like this!

extension/commands/openTransactionViewCommand.ts
```diff
@@ -27,7 +27,9 @@ import { GlobalState } from '../util/GlobalState';
 export async function openTransactionView(treeItem?: InstantiatedTreeItem): Promise<void> {
     const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
     outputAdapter.log(LogType.INFO, undefined, `Open Transaction View`);
-    let smartContract: string;
+    let smartContractLabel: string;
+    let contract: { name: string, contractInstance: {}, transactions: Array<any>, info: {} };
+    let data: { name: string, version: string, channel: string, label: string, transactions: Array<any>, namespace: string };
```
Jakeeyturner: If this is the same as `ISmartContract`, could we create the interface in the extension and then copy the interface file into the `src` directory on build? (Like we do with ExtensionCommands.ts in the build npm script)

PR #1618, opened 2019-11-08
Ops Tools - Save apiKey onto key chain and URL on registry

### LINKED ISSUE #1539: Securely save API key and URL

The method implemented in this epic will not be supported for IBP, since the key only lasts one hour.
Therefore we should not be storing this information.

However in order to test the code we will be using an IBM cloud, which means the restriction will be added at a later point.

toggle007:
Added functionality to allow API keys & url to be stored along with relevant tests
Closes #1539 

### DIFF
--- a/extension/commands/importNodesToEnvironmentCommand.ts
+++ b/extension/commands/importNodesToEnvironmentCommand.ts
@@ -25,6 +25,7 @@ import { FabricEnvironment } from '../fabric/FabricEnvironment';
 import { ExtensionCommands } from '../../ExtensionCommands';
 import { FileSystemUtil } from '../util/FileSystemUtil';
 import Axios from 'axios';
+import { ExtensionUtil } from '../util/ExtensionUtil';
 
 export async function importNodesToEnvironment(environmentRegistryEntry: FabricEnvironmentRegistryEntry, fromAddEnvironment: boolean = false): Promise<boolean> {
     const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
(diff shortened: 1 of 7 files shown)


lquintai:
Looks good.
Two comments to be addressed at your discretion:
1. There are several stubs used in multiple tests being defined within the test. These definitions could be moved to the describe section.
2. Before delivery need to check if this is the right approach to deliver the package-lock.json changes.

lquintai:
it's being stored as an array, so the user could have multiple environments connecting to Ops Tools.
As for the location, seemed a logical place to put it, but we can look into the environment registry.

toggle007:
> Why are you saving the url to the global state? Surely the url should just be saved in the environment registry. We should adda new optional property on an environment for the url?

Have changed the code so theres now an added optional property for URL and adds to the registryentry instead of global state

extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -88,24 +92,44 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
...
                 const api: string = url.replace(/\/$/, '') + GET_ALL_COMPONENTS;
                 const response: any = await Axios.get(api, {
                     headers: { Authorization: `Bearer ${apiKey}` }
                 });
 
+                try {
+                    await keytar.setPassword('blockchain-vscode-ext', url, apiKey);
+                } catch (error) {
+                    outputAdapter.log(LogType.ERROR, `Unable to store API key securely in your keychain: ${error.message}`, `Unable to store API key securely in your keychain: ${error.toString()}`);
+                    return;
```
cazfletch: you need to throw this error so that it gets caught at the bottom so it handles it correctly if it was called from adding an environment. If you just return then add environment will assume it was successful

extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -185,3 +209,17 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
         }
     }
 }
+
+function getCoreNodeModule(moduleName: string): any {
+    try {
+      return ExtensionUtil.getModuleAsar(moduleName);
+    } catch (err) {
+        // do nothing
```
cazfletch: we should tell the user why it failed

extension/extension.ts
```diff
@@ -48,7 +48,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         generatorVersion: originalExtensionData.generatorVersion,
         preReqPageShown: originalExtensionData.preReqPageShown,
         dockerForWindows: originalExtensionData.dockerForWindows,
-        systemRequirements: originalExtensionData.systemRequirements
+        systemRequirements: originalExtensionData.systemRequirements,
```
cazfletch: you don't need this comma

PR #1651, opened 2019-11-18
Ask user to select which nodes should be imported from Ops Tool

### LINKED ISSUE #1543: Ask the user to select the required nodes

After retrieving the Json file we excluded any node types VSCode doesn't handle.

We should then present this list to the user, and let her decide which of the nodes should be used, and then use only those.

Implementation suggestion is to show a list of all the available nodes (that VSCode can handle) with a tickbox next to each node.

lquintai:
Closes #1543

### DIFF
--- a/extension/commands/UserInputUtil.ts
+++ b/extension/commands/UserInputUtil.ts
@@ -974,7 +974,7 @@ export class UserInputUtil {
         return vscode.window.showQuickPick(quickPickItems, quickPickOptions);
     }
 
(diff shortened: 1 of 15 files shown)


extension/commands/UserInputUtil.ts
```diff
@@ -1100,6 +1100,24 @@ export class UserInputUtil {
 
     }
 
+    public static async showOpsToolNodesQuickPickBox(prompt: string, nodes: FabricNode[], canPickMany: boolean): Promise<Array<IBlockchainQuickPickItem<FabricNode>> | IBlockchainQuickPickItem<FabricNode> | undefined> {
```
cazfletch: can we change the name of this function as although you use it for ops tools its not specific to ops tools. I know there is a similar function for choosing nodes, so maybe rename them both, like `showNodesIn Environment` and `showNodes`?

lquintai: All done.

PR #1680, opened 2019-11-22
Add 'hidden' field to FabricNode

### LINKED ISSUE #1542: Save a record of the nodes being used

After retrieving the Json file we excluded any node types VSCode doesn't handle. The user might also decide to not include some of the nodes.

We will write all nodes to disk, but add a new field to the json files so that we have an easily accessible record or which nodes the user added to the environment, out of all the possible nodes in the network.

This record can be used for future reference (when we attempt to reconnect, or when the user wants to add more nodes).

lquintai:
This new field will allow the extension to write all nodes to file, while only presenting to the client the non-hidden nodes.

All nodes are visible by default (hidden = false).

Closes #1542 

### DIFF
--- a/packages/blockchain-extension/extension/fabric/FabricEnvironment.ts
+++ b/packages/blockchain-extension/extension/fabric/FabricEnvironment.ts
@@ -52,13 +52,16 @@ export class FabricEnvironment extends EventEmitter {
         return Array.from(mspIDs).sort();
     }
 
-    public async getNodes(withoutIdentities: boolean = false): Promise<FabricNode[]> {
+    public async getNodes(withoutIdentities: boolean = false, showAll: boolean = false): Promise<FabricNode[]> {
         const rootNodesPath: string = path.resolve(this.path, 'nodes');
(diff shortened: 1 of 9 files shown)


packages/blockchain-extension/extension/fabric/FabricEnvironment.ts
```diff
@@ -55,10 +55,13 @@ export class FabricEnvironment extends EventEmitter {
     public async getNodes(withoutIdentities: boolean = false): Promise<FabricNode[]> {
         const rootNodesPath: string = path.resolve(this.path, 'nodes');
         const nodes: FabricNode[] = await this.loadNodes(rootNodesPath);
+        const visibleNodes: FabricNode[] = nodes.filter((_node: FabricNode) => {
```
cazfletch: do we need to pass into this function whether you want all the nodes or just the visible ones?

lquintai: Good call, we will probably need it in the future. Will implement that.

Issue #741, opened 2019-03-20
support fabric 2


cazfletch:
Here is the broad approach to supporting Fabric 2.0

1.  We’ll start a new GitHub branch of the extension that’s just for 2.0 

2. Therefore there will be a branch of the extension for Fabric v1 and a branch for v2

3. The v1 branch will remain the one “released to the Marketplace” to start with

4. Meanwhile, we’ll be getting the v2 branch all spick and span, and encouraging anyone who wants to try it out (we’ll have a VSIX available from Git)

5. At some point, when we’re fairly confident the v2 one is #goodEnough we’ll “flip the switch” and have the v2 branch be the one that publishes to Marketplace

6. At that point, anyone who wants to continue working with Fabric v1 will need to explicitly install an older version of the extension

7. Any serious issues  found with the v1 dev tools, we can fix in that branch and guide them to install from VSIX to get a version with the fix

caroline-church:
Tried a 2.0 version of fabric with 1.4 images and this all worked ok. I also spoke to Andy Coleman who mentioned that the wallet format might change which could cause a problem with using 1.4 wallets with 2.0 but this isn't a definite thing that will be done. The only other problem could be that the format of errors but this might not go in either

cazfletch:
Questions
 - [ ] Will fabric network/ fabric client/ fabric ca 2.0 support talking to a 1.4 runtime
 - [ ] is there a way to tell whether a fabric is using the new life cycle or old life cycle from the sdk?
          https://fabric-sdk-node.github.io/master/Channel.html#getChannelConfig__anchor is how to get it but you have to decode the block info
- [ ] if i have the Application capabilities at 1.4 and you instantiate some chaincode, then you upgrade the config to have the capabilities set at 2.0. Would i call `getNamspaceDefinitions()` to get the chaincode or `queryInstantiatedChaincode()`
    - Answer from Alessandro - you still need to query the lifecycle that was used to install/instantiate the chaincode in the first place, but the idea is that as you upgrade to 2.0, you should also re-define your chaincodes in the new lifecycle, at which point, you can direct all queries to the new lifecycle

EdMoffatt:
Gonna split design work between myself and Charlotte - here's a list to track what we're working on, and we'll update with designs as we have them!

Things needed to make local fabric support fabric 2.0
- [ ] Update generator to generate projects at either version 1.4 or 2.0 (CM)
[...]

PR #1690, opened 2019-11-26
Create gateway module

cazfletch:
Added a gateway module so that we can have gateways at multiple versions

contributes to #741


packages/blockchain-common/package.json
```diff
@@ -0,0 +1,153 @@
+{
+    "name": "ibm-blockchain-platform-common",
```
Jakeeyturner: Probably worth stripping out all of the key/vals apart from the ones we actually need, as it's just a duplicate of `blockchain-extension`. Also, it's not going to be published to npm either.

Are we able to move the license checker and anything else to root level package.json also?

packages/blockchain-common/src/util/Util.ts
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+'use strict';
+
+import * as fs from 'fs-extra';
+import * as yaml from 'js-yaml';
+
+export class Util {
```
Jakeeyturner: Could we either rename this to something less generic, or move the function elsewhere.
Don't want to have a file like this being used as a dump!

packages/blockchain-common/package.json
cazfletch: The licence checker needs to be in each package

packages/blockchain-extension/cucumber/steps/gateway.steps.ts
```diff
@@ -152,7 +151,7 @@ module.exports = function(): any {
                 openFileNameArray.includes(filePath).should.be.true;
 
                 // Get the smart contract metadata
-                const connection: IFabricClientConnection = FabricConnectionManager.instance().getConnection();
+                const connection: IFabricGatewayConnection = FabricConnectionManager.instance().getConnection();
```
Jakeeyturner: Is FabricConnectionManager ever going to return anything other than a (I)FabricGatewayConnection?
If not - does it make sense to rename FabricConnectionManager to FabricGatewayManager?

packages/blockchain-extension/extension/commands/gatewayConnectCommand.ts
```diff
@@ -124,15 +123,14 @@ export async function gatewayConnect(gatewayRegistryEntry: FabricGatewayRegistry
 
     const connectionProfilePath: string = await FabricGatewayHelper.getConnectionProfilePath(gatewayRegistryEntry.name);
 
-    const connection: IFabricClientConnection = FabricConnectionFactory.createFabricClientConnection(connectionProfilePath);
+    const connection: IFabricGatewayConnection = FabricConnectionFactory.createFabricGatewayConnection(connectionProfilePath);
```
Jakeeyturner: Similar to my last comment - maybe rename FabricConnectionFactory to FabricGatewayFactory?

packages/blockchain-extension/extension/dependencies/DependencyManager.ts
```diff
@@ -530,16 +530,36 @@ export class DependencyManager {
...
+
+                    for (const other of otherGRPC) {
+                        basePath = path.join(extensionPath, 'node_modules', `${other}`, 'node_modules', 'grpc', 'src', 'node', 'extension_binary');
+
+                        origPath = path.join(basePath, `node-v${info.modules}-${os}-${architecture}-${thing}`);
+                        newPath = path.join(basePath, `electron-v${info.shortVersion}-${os}-${architecture}-${thing}`);
+
+                        exists = await fs.pathExists(origPath);
+                        if (exists) {
+                            await fs.remove(origPath);
```
Jakeeyturner: Is it worth try/catching this, just in case?

packages/blockchain-extension/extension/commands/gatewayConnectCommand.ts
cazfletch: that handles the environment connection too so doesn't need to change

Jakeeyturner: Ignore this - just noticed we have FabricEnvironmentConnections in this class!

Jakeeyturner: Beat me to it by 16 seconds!

packages/blockchain-extension/extension/dependencies/DependencyManager.ts
cazfletch: it will get caught be the failed to activate extension bit. The same as the other part that checks if the files exist in the main extension bit. The new code will only ever happen in development, the files will never exist in prod

packages/blockchain-extension/test/commands/gatewayConnectCommand.test.ts
```diff
@@ -73,7 +71,7 @@ describe('GatewayConnectCommand', () => {
         let choseIdentityQuickPick: sinon.SinonStub;
         let chosenGatewayQuickPick: sinon.SinonStub;
         let chosenWalletQuickPick: sinon.SinonStub;
-        let identity: IdentityInfo;
+        let identity: any;
```
Jakeeyturner: Was there a reason for changing this?

cazfletch: as otherwise you need to import `fabric-client` and we need to not do that in the extension package

packages/blockchain-gateway-v1/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
```
Jakeeyturner: Maybe for our tsconfigs, we could extend a base config.
All the settings should be the same across packages, however the includes/excludes may be slightly different.

(I also checked but we can't do this for gitignores unfortunately)

packages/blockchain-gateway-v1/tslint.json
```diff
@@ -0,0 +1,70 @@
+{
```
Jakeeyturner: Would be nice if we could extend a base tslint file also

packages/blockchain-common/test/util/Util.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as chai from 'chai';
+import * as sinon from 'sinon';
+import * as sinonChai from 'sinon-chai';
+import * as chaiAsPromised from 'chai-as-promised';
+
+chai.should();
+chai.use(sinonChai);
+chai.use(chaiAsPromised);
+// tslint:disable no-unused-expression
+describe('Util', () => {
```
Jakeeyturner: This file needs renaming, also `describe('Util', ...`

PR #1691, opened 2019-11-26
Ask for api key and secret when importing nodes from Ops Tools

### LINKED ISSUE #1684: Connect to Ops Tools using API key and secret

For software support we will need to store API key and secret to connect using basic auth  (instead of the key + url initially used in first development stage)

lquintai:
closes #1684 

User will be asked to provide a url, api key and api secret.

If unable to connect, user will be asked if they want to provide a CA certificate chain or to proceed insecurely.

Also fixes the orderer selection: if the user selects any orderer node, all orderer nodes will have hidden = false. 

### DIFF
--- a/packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
+++ b/packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
@@ -18,9 +18,10 @@ import { VSCodeBlockchainOutputAdapter } from '../logging/VSCodeBlockchainOutput
 import { LogType } from '../logging/OutputAdapter';
 import { FabricEnvironmentRegistryEntry } from '../registries/FabricEnvironmentRegistryEntry';
 import * as fs from 'fs-extra';
+import * as https from 'https';
 import * as path from 'path';
 import { SettingConfigurations } from '../../configurations';
-import { FabricNode } from '../fabric/FabricNode';
+import { FabricNode, FabricNodeType } from '../fabric/FabricNode';
 import { FabricEnvironment } from '../fabric/FabricEnvironment';
 import { ExtensionCommands } from '../../ExtensionCommands';
 import { FileSystemUtil } from '../util/FileSystemUtil';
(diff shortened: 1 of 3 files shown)


packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -20,12 +20,13 @@ import { FabricEnvironmentRegistryEntry } from '../registries/FabricEnvironmentR
...
 import * as path from 'path';
 import { SettingConfigurations } from '../../configurations';
-import { FabricNode } from '../fabric/FabricNode';
+import { FabricNode, FabricNodeType } from '../fabric/FabricNode';
 import { FabricEnvironment } from '../fabric/FabricEnvironment';
 import { ExtensionCommands } from '../../ExtensionCommands';
 import { FileSystemUtil } from '../util/FileSystemUtil';
 import Axios from 'axios';
 import { ExtensionUtil } from '../util/ExtensionUtil';
+import https = require('https');
```
cazfletch: this should be imported like other node modules

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -117,13 +117,51 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
...
+                            const browseOptions: vscode.OpenDialogOptions = {
+                                canSelectFiles: true,
+                                canSelectFolders: false,
+                                canSelectMany: false,
+                                openLabel: 'Select',
+                                filters: {
+                                    Certificates: ['pem']
+                                }
+                            };
+                            const certificatePath: vscode.Uri = await UserInputUtil.browse('Enter a file path to the collection configuration', quickPickItems, browseOptions, true) as vscode.Uri;
```
toggle007: Perhaps we should change the message shown to the user (as they aren't entering a file path, but choosing a file instead?) to something like:
`Select the certificate (pem) you wish to add`

lquintai: Well spotted :)

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
lquintai: Addressed.

PR #1735, opened 2019-12-03
Ops tools2 save ca cert

### LINKED ISSUE #1719: Ops Tools - Copy CA certificate chain to the environment base directory
If a CA certificate chain is provided, we need to store it so that we don't ask for it again on reconnect/Edit filters/refresh.

This file should be stored on the environment base directory (one up from nodes).

lquintai:
Save CA certificate chain file in the environment base directory when provided by the user.

closes #1719

### DIFF
--- a/packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
+++ b/packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
@@ -24,6 +24,8 @@ import { ExtensionCommands } from '../../ExtensionCommands';
 
 export async function addEnvironment(): Promise<void> {
     const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+    const fabricEnvironmentEntry: FabricEnvironmentRegistryEntry = new FabricEnvironmentRegistryEntry();
+    const fabricEnvironmentRegistry: FabricEnvironmentRegistry = FabricEnvironmentRegistry.instance();
     try {
         outputAdapter.log(LogType.INFO, undefined, 'Add environment');
 
@@ -31,8 +33,6 @@ export async function addEnvironment(): Promise<void> {
         if (!createMethod) {
             return;
         }
-
-        const fabricEnvironmentRegistry: FabricEnvironmentRegistry = FabricEnvironmentRegistry.instance();
 
         const environmentName: string = await UserInputUtil.showInputBox('Enter a name for the environment');
         if (!environmentName) {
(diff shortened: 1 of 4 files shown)


packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -221,11 +235,10 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
 
         const dirPath: string = await vscode.workspace.getConfiguration().get(SettingConfigurations.EXTENSION_DIRECTORY) as string;
         const homeExtDir: string = FileSystemUtil.getDirPath(dirPath);
-        const environmentPath: string = path.join(homeExtDir, 'environments', environmentRegistryEntry.name, 'nodes');
+        const environmentNodesPath: string = path.join(homeExtDir, 'environments', environmentRegistryEntry.name, 'nodes');
```
cazfletch: you don't need all this you can use your variable you have at the top for the `environmentBaseDir`

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -251,7 +264,7 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
         const newEnvironment: FabricEnvironment = new FabricEnvironment(environmentRegistryEntry.name);
         const newNodes: FabricNode[] = await newEnvironment.getNodes();
         if (newNodes.length === 0) {
-            await fs.remove(environmentPath);
+            await fs.remove(environmentNodesPath);
```
cazfletch: in the case of a non ops tools one this needs to remove the environment

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
lquintai: True, will refactor.

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
lquintai: Indeed, the behavior is now different from what is currently expected, I'll change that.

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -62,6 +61,9 @@ export async function addEnvironment(): Promise<void> {
         }
         Reporter.instance().sendTelemetryEvent('addEnvironmentCommand');
     } catch (error) {
+        if (fabricEnvironmentEntry.name) {
```
cazfletch: You don't need this bit as the delete will ignore the error if it doesn't exist

PR #1741, opened 2019-12-06
Added functional transaction panel to transaction view

### LINKED ISSUE #1488: As a user I can see any output from my transactions
Output will appear in the panel on the right hand side of the create transaction view as seen below. When there is no output present, it should just show a generic message. After a transaction has been submitted/evaluated, the output we get back from Fabric should be displayed in this panel instead.

image

erin-hughes:
Closes #1488 and contributes to #1155 

Signed-off-by: Erin Hughes Erin.Hughes@ibm.com

### DIFF
--- a/packages/blockchain-ui/enzyme/tests/TransactionOutput.test.tsx
+++ b/packages/blockchain-ui/enzyme/tests/TransactionOutput.test.tsx
@@ -1,0 +1,17 @@
+// tslint:disable no-unused-expression
+import React from 'react';
+import renderer from 'react-test-renderer';
+import TransactionOutput from '../../src/components/TransactionOutput/TransactionOutput';
+import chai from 'chai';
+import sinonChai from 'sinon-chai';
+chai.should();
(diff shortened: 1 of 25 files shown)


packages/blockchain-ui/enzyme/tests/TransactionOutput.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+import renderer from 'react-test-renderer';
+import TransactionOutput from '../../src/components/TransactionOutput/TransactionOutput';
+import chai from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+chai.should();
+chai.use(sinonChai);
+
+describe('TransactionOutput component', () => {
+    let mySandBox: sinon.SinonSandbox;
```
Jakeeyturner: Probably don't need any of the sandbox stuff in this file (ignore this if this will change in a future PR though).

PR #1760, opened 2019-12-12
Update debug configurations

### LINKED ISSUE #1079: Cannot debug chaincode created not using vscode extension
There are probably two related but separate issues in the description below.

We wrote a chaincode component in TypeScript from scratch, meaning we did **not** use the capability provided by the vscode extension to generate the scaffolding code.

Though we are able to use the vscode extension to package, install, and instantiate this chaincode component, we cannot debug it. If you go to the debug panel in vscode, you do not get the `Debug Smart Contract` option. Instead, you see the `No Configurations` option and the option to add a new configuration. Choosing the option for adding a new configuration offers then several options for Fabric smart contracts. However, choosing the option for Fabric node generates a `launch.json` file that does not yield the desired outcome... meaning, you cannot still debug the chaincode component.

We then decided to copy the `launch.json` file that was created in another chaincode project (one that was generated using the vscode extension) and brought that `launch.json` file over to the chaincode project that was written from scratch (please note that this chaincode component happens to use the low level chaincode api). After that, then we can see `Debug Smart Contract` configuration/option in the debugger view... but starting it fails because it attempts to find `fabric-contract-api` module. Therefore, it looks like we need a different debug configuration that does not depend on the `fabric-contract-api` module so that we can debug chaincode written using the low level api.

## Your Environment
* IBM Blockchain Platform Extension Version: 1.0.3
* Visual Studio Code version: 1.35.1
* Environment name and version (e.g. Node.js v8.12.0, npm v6.4.1): node v8.9.4, npm 6.9.0
* Operating System and version: macOS (Mojave 10.14.5)

cazfletch:
Updated the debug configurations so that they match what is generated by the generator

closes #1079


Jakeeyturner:
Tested it out and it seems ok - will approve and merge once our release it out.

PR #1775, opened 2019-12-17
create ops tool environment without nodes closes #1721

### LINKED ISSUE #1721: Ops Tools - Environment can be created without nodes and show as disconnected
When creating the Ops Tools environment, the user can decide not to choose any nodes.

If no nodes are chosen, then we should allow the process to continue, ie, we should store all the necessary information (at this point we know the connection was successful), and the environment should be created.

The toast notification for successfully creating the environment should read: 'No available nodes included in current filters. Click ENVIRONMENT_NAME to edit filters'

Environment should show as disconnected.

lquintai:
closes #1721  

The extra functionality needed for the correct behavior, ie., not allowing to connect until there are nodes in the environment, will be delivered in #1775 

### DIFF
--- a/packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
+++ b/packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
@@ -20,7 +20,7 @@ import * as vscode from 'vscode';
 import { UserInputUtil } from './UserInputUtil';
 import { Reporter } from '../util/Reporter';
 import { VSCodeBlockchainOutputAdapter } from '../logging/VSCodeBlockchainOutputAdapter';
-import { FabricEnvironmentRegistry, FabricEnvironmentRegistryEntry, FabricRuntimeUtil, LogType } from 'ibm-blockchain-platform-common';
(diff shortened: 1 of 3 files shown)


packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -135,18 +135,21 @@ export async function addEnvironment(): Promise<void> {
...
-
         if (addedAllNodes === undefined) {
             await fabricEnvironmentRegistry.delete(fabricEnvironmentEntry.name);
             return;
-        }
-
-        if (addedAllNodes) {
-            outputAdapter.log(LogType.SUCCESS, 'Successfully added a new environment');
+        } else if (addedAllNodes) {
+            const environment: FabricEnvironment = new FabricEnvironment(fabricEnvironmentEntry.name);
```
cazfletch: If the nodes are added from an environment then you shouldn't let it create with no nodes, only if its from ops tools

lquintai: If no nodes are added in a non-OpsTools environment in the importNodesToEnvironment command, this will result in an error being thrown, ie, it is only possible to be in this code path if it is an opsTool environment.

PR #1792, opened 2019-12-27
Docker build fix

au-shah:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-FROM codercom/code-server:v2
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-FROM codercom/code-server:v2
+FROM codercom/code-server:v2.1699-vsc1.41.0
```
cazfletch: this line should be `FROM codercom/code-server:2.1699-vsc1.41.0` you have a v before the version

au-shah: Fixed

Issue #1154, opened 2019-07-16
Update tutorial to show short cut for instantiate


cazfletch:
Update first tutorial to use short cut way of instantiating and say you can also do it the long way

EdMoffatt:
Adding this to the current sprint (I'll make the changes and submit a PR for this one!)

EdMoffatt:
Had a chat with MGK and here's the planned set of updates to the tutorial:

1. Replace 3 step process with the "quick" method
2. Add a step that has them make a minor change to the smart contract and package.json
3. Show an upgrade, using the "slow" method

That way we show both but we don't offer 2 ways of doing the same thing (which MGK reckons is highly likely to confuse people in labs!)

cazfletch:
done so closing

PR #1796, opened 2020-01-02
Update local-dev.md

EdMoffatt:
Updating the first tutorial with some minor fixes. The most significant being:

1. Add the "choose endorsement policy" to the instantiate flow

2. Fix issue #1154 by restructuring to recommend the single-step instantiate method and having the 3-steps as an alternative.

Suggest we get MGK to review this PR before merging, since it will effect his labs! I'll drop him an email about it.


EdMoffatt:
Closing because MGK had some good suggestions (via email to me) about how we could structure this to avoid confusion in a lab!

PR #1840, opened 2020-01-22
Ops tools3 filter nodes

### LINKED ISSUE #1546: Add node/'Edit filters' to a network imported from Ops Tool, inform user of changes and provide option to modify nodes being used

- retrieve stored API key + secret
- connect to Ops Tool
- import Json

How:
- filter out nodes VSCode doesn't understand
- compare list of nodes retrieved with the list of nodes present.
- present a checkbox list of all the nodes available:
  - show the nodes currently in use already ticked and at the bottom of the list
  - show the nodes that already existed but were hidden unticked, in the middle of the list
  - show new nodes at the top of the list, with '(new)' after the name

This option to modify the set of nodes being used will be shared by reconnect, refresh, add node and automatic scans of ops tools. This means this work will be to connect the infrastructure to the reconnect action.

A further issue will cover creating the toast notification that will alert the user, and only if the user chooses to edit the filters then this list will appear.

**Checklist**

- [ ] Show all nodes to the user so they can pick using a tickbox

- [ ] Tick all nodes that are known & _not_ hidden

- [ ] Add description to newly discovered nodes

lquintai:
Closes #1546

### DIFF
--- a/packages/blockchain-extension/extension/commands/UserInputUtil.ts
+++ b/packages/blockchain-extension/extension/commands/UserInputUtil.ts
@@ -1043,23 +1043,30 @@ export class UserInputUtil {
         });
     }
 
-    public static async showNodesQuickPickBox(prompt: string, nodes: FabricNode[], canPickMany: boolean, nonHiddenNodes?: FabricNode[]): Promise<Array<IBlockchainQuickPickItem<FabricNode>> | IBlockchainQuickPickItem<FabricNode> | undefined> {
+    public static async showNodesQuickPickBox(prompt: string, nodes: FabricNode[], canPickMany: boolean, currentNodes?: FabricNode[]): Promise<Array<IBlockchainQuickPickItem<FabricNode>> | IBlockchainQuickPickItem<FabricNode> | undefined> {
         if (nodes.length === 0) {
             throw new Error('Error when importing nodes, no nodes found to choose from.');
         }
 
         const quickPickItems: IBlockchainQuickPickItem<FabricNode>[] = UserInputUtil.selectNodesOneOrdererPerCluster(nodes);
 
-        if (nonHiddenNodes) {
+        if (currentNodes && currentNodes.length > 0) {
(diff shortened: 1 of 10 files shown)


packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -26,6 +26,7 @@ import { ModuleUtil } from '../util/ModuleUtil';
 
 export async function importNodesToEnvironment(environmentRegistryEntry: FabricEnvironmentRegistryEntry, fromAddEnvironment: boolean = false, createMethod?: string): Promise<boolean> {
     const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+    const text: string = createMethod !== UserInputUtil.ADD_ENVIRONMENT_FROM_OPS_TOOLS ? 'import' : 'filter';
```
au-shah: Maybe this variable should be named better? So we know what text actually is?
Something like createMethodString?

packages/blockchain-extension/extension/commands/UserInputUtil.ts
```diff
@@ -1043,23 +1043,55 @@ export class UserInputUtil {
...
         if (nodes.length === 0) {
             throw new Error('Error when importing nodes, no nodes found to choose from.');
         }
 
         const quickPickItems: IBlockchainQuickPickItem<FabricNode>[] = UserInputUtil.selectNodesOneOrdererPerCluster(nodes);
 
+        if (currentNodes && currentNodes.length > 0) {
+            const currentUrls: string[] = currentNodes.map((_node: FabricNode) => _node.api_url);
+            quickPickItems.map((quickPickItem: IBlockchainQuickPickItem<FabricNode>) => {
+                currentNodes.forEach((node: FabricNode) => {
```
cazfletch: i'm not sure a forEach is the correct thing to use here. Maybe a find to get the correct node and if it doesn't find it you know it's a new one

PR #1845, opened 2020-01-27
Edit filters - reconnect. Ask if user wants to edit filters when diff…

### LINKED ISSUE #1544: On reconnect, inform user of changes and provide option to modify nodes being used

- retrieve stored API key + secret
- connect to Ops Tool
- import Json

How:
- filter out nodes VSCode doesn't understand
- compare list of nodes retrieved with the list of nodes present.
- present a checkbox list of all the nodes available:
  - show the nodes currently in use already ticked and at the bottom of the list
  - show the nodes that already existed but were hidden unticked, in the middle of the list
  - show new nodes at the top of the list, with '(new)' after the name

This option to modify the set of nodes being used will be shared by reconnect, refresh, add node and automatic scans of ops tools. This means this work will be to connect the infrastructure to the reconnect action.

A further issue will cover creating the toast notification that will alert the user, and only if the user chooses to edit the filters then this list will appear.

lquintai:
…erences detected. closes #1544


packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
```diff
@@ -56,6 +56,24 @@ export async function fabricEnvironmentConnect(fabricEnvironmentRegistryEntry: F
...
+        let nodes: FabricNode[] = await fabricEnvironment.getNodes();
+
+        if (fabricEnvironmentRegistryEntry.url) {
+            let informOfChanges: boolean = true;
+            if (nodes.length === 0) {
+                const importNodes: boolean = await UserInputUtil.showConfirmationWarningMessage(`Error connecting to environment ${fabricEnvironmentRegistryEntry.name}: no visible nodes. Would you like to filter nodes?`);
+                if (!importNodes) {
+                    return;
+                }
+                informOfChanges = false;
```
cazfletch: why do you set this back to false?

packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
```diff
@@ -56,6 +56,24 @@ export async function fabricEnvironmentConnect(fabricEnvironmentRegistryEntry: F
...
+            }
+            await vscode.commands.executeCommand(ExtensionCommands.EDIT_NODE_FILTERS, fabricEnvironmentRegistryEntry, false, UserInputUtil.ADD_ENVIRONMENT_FROM_OPS_TOOLS, informOfChanges);
+            nodes = await fabricEnvironment.getNodes();
+            if (nodes.length === 0) {
+                return;
```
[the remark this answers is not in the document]
lquintai: Not at this point, because this PR deals with reconnect, so we are coming from a disconnected scenario. Next PR will deal with refresh, ie, connect when already connected. This will be addressed then.

packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
lquintai: At this stage we already asked if the user wants to edit filters, and the answer was yes. We don't want to ask again (which we would if differences were detected and informOfChanges was true) because we already decided to show the tickbox.
The tickbox will display any differences as part of the node's description (<new>/<was OLD NAME>) so the user will not miss out on information..

cazfletch: ahhh i see i missed where it was being used again

PR #1948, opened 2020-02-18
Update generator to generate txdata files in smart contract projects

### LINKED ISSUE #1871: Make the generator generate transaction data
For each of the contracts we should generate transaction data

erin-hughes:
Also added Cucumber tests for txdata

Closes #1871 

### DIFF
--- a/packages/blockchain-extension/package.json
+++ b/packages/blockchain-extension/package.json
@@ -1333,7 +1333,7 @@
         "ejs": "^2.6.1",
         "find-free-port": "^2.0.0",
         "fs-extra": "^7.0.1",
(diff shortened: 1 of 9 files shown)


packages/blockchain-extension/package.json
```diff
@@ -1333,7 +1333,7 @@
         "ejs": "^2.6.1",
         "find-free-port": "^2.0.0",
         "fs-extra": "^7.0.1",
-        "generator-fabric": "0.1.6",
+        "generator-fabric": "^0.1.7",
```
cazfletch: We shouldn't have the `^` as we need to be sure of which version

PR #1976, opened 2020-02-24
User must now select the channel in which the smart contract they wish to submit/evaluate is

### LINKED ISSUE #1777: Submit/Evaluate transaction flow should ask for channel
When doing a submit or evaluate via the Command Palette, you're asked to select:

- Gateway
- Identity
- Smart Contract

Because you can have a smart contract of the same name and version instantiated on 2 channels, that means you can end up selecting from a list like this:

image

I think we need to add in a step of selecting your channel in between the Identity and Smart Contract steps.

au-shah:
closes #1777 

### DIFF
--- a/packages/blockchain-extension/extension/commands/UserInputUtil.ts
+++ b/packages/blockchain-extension/extension/commands/UserInputUtil.ts
@@ -464,6 +464,34 @@ export class UserInputUtil {
         }
     }
 
+    public static async showChannelFromGatewayQuickPickBox(prompt: string): Promise<IBlockchainQuickPickItem<Array<string>>> {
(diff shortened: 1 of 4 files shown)


packages/blockchain-extension/extension/commands/UserInputUtil.ts
```diff
@@ -464,6 +464,26 @@ export class UserInputUtil {
         }
     }
 
+    public static async showChannelFromGatewayQuickPickBox(prompt: string, channelMap: Map<string, Array<string>>): Promise<IBlockchainQuickPickItem<Array<string>>> {
```
cazfletch: you should move the getConnection and createChannelMap part in here

Issue #2069, opened 2020-03-19
Confirmation message on generator update appears if there are no local environments


Jakeeyturner:
## Description

When the version of the generator updates (in a new release), if the user hasn't got any local environments/runtimes, we still ask if they want to teardown their local runtimes to update.

We shouldn't do this as we can just update the global state automatically.

I've found the problem in the code, so can fix it!
## Steps to Reproduce

1.
2.
3.

## Existing issues

- [ ] [Stack Overflow issues](https://stackoverflow.com/questions/tagged/ibp-vscode-extension)
- [ ] [GitHub Issues](https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues)

## Your Environment

* IBM Blockchain Platform Extension Version:
* Visual Studio Code version:
* Environment name and version (e.g. Node.js v8.12.0, npm v6.4.1):
* Operating System and version:
* Link to your project:

PR #2071, opened 2020-03-19
Hide generator teardown message if there are no local environments (c…

Jakeeyturner:
…lose #2069)


packages/blockchain-extension/extension/util/ExtensionUtil.ts
```diff
@@ -658,64 +658,67 @@ export class ExtensionUtil {
...
-            const generatorMajor: number = generatorSemver.major;
-            const storedMajor: number = storedSemver.major;
-            if (generatorMajor > storedMajor) {
-                // If major changes, then we should just force teardown without asking.
-                teardownRuntimes = true;
-            }
+                const generatorMajor: number = generatorSemver.major;
+                const storedMajor: number = storedSemver.major;
+                if (generatorMajor > storedMajor) {
+                    // If major changes, then we should just force teardown without asking.
```
cazfletch: I think we need to tell the user we will tear them down and why

Jakeeyturner: If there is a major version, `teardownRuntimes = true`, then on line 685, we will log that all local runtimes will be torn down.

PR #2093, opened 2020-03-27
IBM OpsTools - add environment. Closes #1339

### LINKED ISSUE #1339: As a user i can add an environment that will discover from ibm cloud
When adding an environment there should be the option of discover from ibm cloud this will let you log into ibm cloud and then find the node files

lquintai:

### DIFF
--- a/packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
+++ b/packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
@@ -28,6 +28,8 @@ import { LocalEnvironmentManager } from '../fabric/environments/LocalEnvironment
 import { SettingConfigurations } from '../../configurations';
 import { ExtensionUtil } from '../util/ExtensionUtil';
 import { GlobalState, ExtensionData } from '../util/GlobalState';
+import { URL } from 'url';
+import { ExtensionsInteractionUtil } from '../util/ExtensionsInteractionUtil';
 
 export async function addEnvironment(): Promise<void> {
     const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
@@ -55,6 +57,7 @@ export async function addEnvironment(): Promise<void> {
         createMethod = chosenMethod.data;
 
         let configurationChosen: number; // Configuration chosen (e.g. 1 Org, 2 Org)
+        let defaultName: string = '';
 
         if (createMethod === UserInputUtil.ADD_ENVIRONMENT_FROM_TEMPLATE) {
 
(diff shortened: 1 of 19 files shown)


packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -94,78 +97,153 @@ export async function addEnvironment(): Promise<void> {
...
-            const HEALTH_CHECK: string = '/ak/api/v1/health';
-            const GET_ALL_COMPONENTS: string = '/ak/api/v1/components';
-            let url: string;
-            const userUrl: string = await UserInputUtil.showInputBox('Enter the URL of the IBM Blockchain Platform Console you want to connect to');
-            if (!userUrl) {
+            const isSaaS: string = await UserInputUtil.showQuickPickYesNo('Are you connecting to a service instance on IBM Cloud?');
+            if (!isSaaS) {
                 return;
+            } else if (isSaaS === UserInputUtil.YES) {
+                let url: string;
```
cazfletch, line 104: can we split some of this file out into functions. The functions can be in this file but its getting to be a really big function

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -128,32 +129,52 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
...
+            // establish connection to Ops Tools
             try {
-                // establish connection to Ops Tools and retrieve json file
-                const credentials: string = await keytar.getPassword('blockchain-vscode-ext', environmentRegistryEntry.url);
-                const credentialsArray: string[] = credentials.split(':');
-                if (credentialsArray.length !== 3) {
-                    throw new Error(`Unable to retrieve the stored credentials`);
+                if (environmentRegistryEntry.environmentType === EnvironmentType.SAAS_OPS_TOOLS_ENVIRONMENT) {
+                    const cloudAccountExtension: vscode.Extension<any> = vscode.extensions.getExtension( 'IBM.ibmcloud-account' );
+                    if ( !cloudAccountExtension.isActive ) {
```
cazfletch, line 141: Can we split this code out into a separate file as it seems duplicated, also why is there less log in code here

packages/blockchain-extension/package.json
```diff
@@ -1504,7 +1505,8 @@
             ".pid",
             ".port",
             ".feature",
-            ".xml"
+            ".xml",
+            ".ts"
```
cazfletch, line 1509: we shouldn't be ignoring .ts files in the licence checker as all typescript files should have a licence

lquintai: indeed, modified license format (it's the same license)

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
lquintai: done. Login code is the same, but more work is needed to get the console endpoint.

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
lquintai: done.

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -286,4 +241,128 @@ export async function addEnvironment(): Promise<void> {
...
+        const ibpResources: IBlockchainQuickPickItem<string>[] = [];
+        const accessToken: string = await ExtensionsInteractionUtil.cloudAccountGetAccessToken(true);
+        if (!accessToken) {
+            return;
+        }
+        const requestOptions: any = { headers: { Authorization: `Bearer ${accessToken}` } };
+        const baseUrl: string = 'https://resource-controller.cloud.ibm.com';
+        let resourcesUrl: string = `${baseUrl}/v2/resource_instances`;
+        while (resourcesUrl) {
+            const response: any = await Axios.get(resourcesUrl.toString(), requestOptions);
```
Jakeeyturner, line 326: Is there a reason / a scenario when resourcesUrl is not a string?

packages/blockchain-extension/extension/cloud-account-api.d.ts
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+*/
+
+import { URL } from "url";
+
+/**
+ * The external interface exposed by the IBM Cloud Account extension
+ * for use by other modules to interact with IBM Cloud APIs.
+ */
+export interface CloudAccountApi {
```
Jakeeyturner, line 21: Might be worth us creating an `interfaces` directory and putting this file in there - just in case we add more in the future.

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -286,4 +241,128 @@ export async function addEnvironment(): Promise<void> {
...
+
+            for (const resource of response.data.resources) {
+                if ( resource.resource_plan_id === 'blockchain-standard' ) {
```
Jakeeyturner, line 329: What is the `resource_plan_id`? Does it come from IBP? Do we not want to support other resource_plan_id's?

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -286,4 +241,128 @@ export async function addEnvironment(): Promise<void> {
...
+            }
+        }
+
+        let chosenIbp: IBlockchainQuickPickItem<any>;
+        if (ibpResources.length === 0) {
+            throw new Error('There are no IBM Blockchain Platform resources associated with the chosen account');
+        } else if (ibpResources.length === 1) {
+            chosenIbp = ibpResources[0];
+        } else {
+            chosenIbp = await UserInputUtil.showQuickPickItem('Please select the IBM Blockchain Platform to connect to', ibpResources) as IBlockchainQuickPickItem<any>;
```
Jakeeyturner, line 350: Is this message clear enough for a user - would `Please select the IBM Blockchain Platform resource to connect to` be clearer?

packages/blockchain-extension/extension/util/ExtensionsInteractionUtil.ts
```diff
@@ -0,0 +1,58 @@
...
+            hasAccount = await cloudAccount.accountSelected();
+            if (!hasAccount) {
+                // If not logged in, this will first ask the user to login, and then to select the account.
+                await vscode.commands.executeCommand('ibmcloud-account.selectAccount');
+            }
+        }
+
+        const isLoggedIn: boolean = await cloudAccount.loggedIn();
+        if (!isLoggedIn) {
+            return; // FIXME - evaluate if an error should be thrown after changes to cloud extension - tell the user we can't proceed without being logged in
```
Jakeeyturner, line 45: Will this be fixed in another PR?

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
lquintai: It's what identifies the resource as an IBM Blockchain platform, as opposed to any other resource in IBM cloud. We don't want to support any other.

packages/blockchain-extension/extension/util/ExtensionsInteractionUtil.ts
lquintai: All the current FIXME will be addressed once there is a decision regarding the cloud login extension modifications I proposed.

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -128,32 +129,44 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
...
+                    }
+                    requestOptions = { headers: { Authorization: `Bearer ${accessToken}` } };
+                } else {
+                    const keytar: any = ModuleUtil.getCoreNodeModule('keytar');
+                    if (!keytar) {
+                        throw new Error('Error importing the keytar module');
+                    }
+
+                    const credentials: string = await keytar.getPassword('blockchain-vscode-ext', environmentRegistryEntry.url);
+                    const credentialsArray: string[] = credentials.split(':');
```
Jakeeyturner, line 152: Just wondering where you've done 'splits', if its worth writing a comment that says what you're expecting the output to be? Might make it easier to debug if Cloud changes the values/format.

E.g. `const credentialsArray: string[] = credentials.split(':'); // Format ['hello', 'world', 'somethingElse']`

Don't know - what do you think?

packages/blockchain-extension/extension/cloud-account-api.d.ts
lquintai: Will do.

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
lquintai: Good catch, I think by then I enforced type already.

packages/blockchain-extension/package.json
```diff
@@ -1341,7 +1341,7 @@
         "pretest": "npm run updatePackageJSON",
         "test": "node ./build/test/runTests.js",
         "posttest": "remap-istanbul -i ./coverage/coverage.json -o ./coverage/coverage-remap.json && istanbul check-coverage ./coverage/coverage-remap.json",
-        "cucumber": "node ./build/cucumber/runCucumberTests.js",
+        "cucumber": "npm run updatePackageJSON && node ./build/cucumber/runCucumberTests.js",
```
Jakeeyturner, line 1344: Why do we need to do this?

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
lquintai: Just checked the design and actually the message should be 'Select an IBM Blockchain Platform service instance', so I'll change it to that.

lquintai: Also changed the wording of the previous error to say 'instances' instead of 'resources'

packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
lquintai: The format is decided by us in 'addEnvironmentCommand', so our extension is in control - there will be no changes unless we make them. Happy to add a comment for readability though.

packages/blockchain-extension/package.json
lquintai: Ah.. Tests failed in azure, and it looked like the changes I made modified activation events slightly. Maybe something to do with activating the new extension dependency? Talked with Caroline, who's had a similar problem, and the suggestion was to updatePackageJSON just before running the cucumber tests, to guarantee activation events happen in the correct order and therefore tests start with a clean slate. This is the solution used in V2.

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -286,4 +241,128 @@ export async function addEnvironment(): Promise<void> {
...
+                if ( resource.resource_plan_id === 'blockchain-standard' ) {
+                    ibpResources.push({
+                        label: resource.name,
+                        description: resource.guid,
+                        data: resource
+                    });
+                }
+            }
+            if (response.data.next_url) {
+                resourcesUrl = `${baseUrl}${response.data.next_url.toString()}`;
```
Jakeeyturner, line 338: Do we need to call .toString() here?

packages/blockchain-extension/extension/commands/addEnvironmentCommand.ts
```diff
@@ -286,4 +241,128 @@ export async function addEnvironment(): Promise<void> {
...
+            chosenIbp = await UserInputUtil.showQuickPickItem('Select an IBM Blockchain Platform service instance', ibpResources) as IBlockchainQuickPickItem<any>;
+            if (!chosenIbp) {
+                return;
+            }
+        }
+
+        // get API endpoint
+        const dashboardUrl: URL = new URL(chosenIbp.data.dashboard_url);
+        const encodedCrn: string = encodeURIComponent(chosenIbp.data.crn);
+        const consoleStatus: any = await Axios.get(`${dashboardUrl.origin.toString()}/api/alternative-auth/resources/${encodedCrn}/optools`, requestOptions);
```
Jakeeyturner, line 359: Do we need to call .toString() here as well?

PR #2115, opened 2020-03-31
Fabric 2: Package a smart contract

### LINKED ISSUE #2113: Fabric 2: As a user i can package a smart contract
I can package a smart contract using the new packaging format

cazfletch:
Added being able to package a smart contract
This doesn't include go packages with a mod file
Commented out all cucumber tests that now won't work

closes #2113


packages/blockchain-extension/extension/commands/exportSmartContractPackageCommand.ts
```diff
@@ -39,7 +39,14 @@ export async function exportSmartContractPackage(packageTreeItem?: PackageTreeIt
             }
             packageToExport = chosenPackage.data;
         }
-        const defaultPath: string = path.join(os.homedir(), `${packageToExport.name}@${packageToExport.version}.cds`);
+
+        let packageName: string = packageToExport.name;
+
+        if (packageToExport.version) {
```
Jakeeyturner, line 45: In v2 it's possible to create a package without a version?

packages/blockchain-extension/extension/commands/importSmartContractPackageCommand.ts
```diff
@@ -31,7 +31,7 @@ export async function importSmartContractPackageCommand(): Promise<void> {
         canSelectFolders: false,
         canSelectMany: false,
         openLabel: 'Select',
-        filters: { Packages: ['cds'] }
+        filters: { Packages: ['tar.gz', 'tgz'] }
```
Jakeeyturner, line 34: If you haven't already, it's worth checking that what we don't expect the file extension to only be `.tar.gz` now that we can provide a `.tar.gz` or `.tgz`.

Jakeeyturner: Will I be able to import both `contract@0.0.1.tar.gz` and `contract@0.0.1.tgz`?

Jakeeyturner: Also, do we validate the name of the imported package? Does it even matter - do we use the package name for anything other than the tree item label?

packages/blockchain-extension/extension/commands/exportSmartContractPackageCommand.ts
cazfletch: yes in v2 packages don't have versions. We are using the version property to name the file but if the package was created outside of vscode then it might not have the same format filename

packages/blockchain-extension/extension/commands/importSmartContractPackageCommand.ts
cazfletch: the package files in the tests have a mixture of tar.gz and .tgz

cazfletch: yeah you can import either, the name of the imported package doesn't matter. We will eventually use it for the definition when you approve

packages/blockchain-extension/extension/commands/importSmartContractPackageCommand.ts
```diff
@@ -49,6 +49,29 @@ export async function importSmartContractPackageCommand(): Promise<void> {
...
 
         resolvedPkgDir = path.join(resolvedPkgDir, packageName);
+
+        let existingPackages: string[] = await fs.readdir(packagePath);
+
+        const nameRegex: RegExp = new RegExp(/(.+?)(\.tar\.gz|\.tgz)/gm);
+
+        const result: RegExpExecArray = nameRegex.exec(packageName);
+
+        const newName: string = result[1];
```
Jakeeyturner, line 59: If the package name isn't matched, then might `result` be of length 0?

cazfletch: it will always be matched as you can only import tar.gz or tgz files

PR #2120, opened 2020-04-01
Making entire tutorials tile on home page clickable

erin-hughes:


Jakeeyturner:
Tried it out and it works as intended - approving!

PR #2143, opened 2020-04-03
Fixed telemetry events

### LINKED ISSUE #2050: Update environment connect telemetry
Now we have different types of environments we can connect to we should update the telemetry for connecting to an environment to show which type of environment a user is connecting to

au-shah:
closes #2050 

### DIFF
--- a/packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
+++ b/packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
@@ -119,8 +119,19 @@ export async function fabricEnvironmentConnect(fabricEnvironmentRegistryEntry: F
             environmentData = 'user environment';
         }
 
+        let envType: string;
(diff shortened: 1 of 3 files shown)


packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
```diff
@@ -119,8 +119,19 @@ export async function fabricEnvironmentConnect(fabricEnvironmentRegistryEntry: F
             environmentData = 'user environment';
         }
 
+        let envType: string;
+        if (fabricEnvironmentRegistryEntry.environmentType === EnvironmentType.ENVIRONMENT) {
```
Jakeeyturner, line 123: Could you comment in `FabricEnvironmentRegistryEntry.ts` (where we've defined the EnvironmentType enum), to state that if we update the enum, we need to update this part here as well, so it gets reported!

au-shah: Done! 👍

Issue #1250, opened 2019-08-06
As a user i can create a gateway to a 2.0 fabric


cazfletch:
Should be able to create a gateway that connects to a 2.0 fabric and is able to submit transactions. This might just involve updating the version of `fabric-network` but we might need to write a connector if we need to use `fabric-network:1.4` and `fabric-network:2.0`

cazfletch:
Can't estimate until know what implementation is: 1 or 5

cazfletch:
done so closing

PR #2192, opened 2020-04-16
Update gateway package

cazfletch:
updated to use new apis

contributes to #1250

### DIFF
--- a/packages/blockchain-gateway-v1/src/FabricConnection.ts
+++ b/packages/blockchain-gateway-v1/src/FabricConnection.ts
@@ -13,11 +13,12 @@
 */
 'use strict';
 
-import * as Client from 'fabric-client';
 import { Gateway, GatewayOptions, Network, Wallet } from 'fabric-network';
 import { URL } from 'url';
 import { ConsoleOutputAdapter, FabricChaincode, OutputAdapter } from 'ibm-blockchain-platform-common';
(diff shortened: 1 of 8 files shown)


packages/blockchain-gateway-v1/src/FabricConnection.ts
```diff
@@ -168,38 +173,29 @@ export abstract class FabricConnection {
...
-        const peers: Client.Peer[] = this.getAllPeers();
-        let lastError: Error = new Error(`Could not discover information for channel ${channelName} from known peers`);
-        for (const target of peers) {
-            try {
-                await channel.initialize({ asLocalhost: this.discoveryAsLocalhost, discover: this.discoveryEnabled, target });
-                return channel;
-            } catch (error) {
-                lastError = error;
-            }
+        // This bit is needed to add all the peers to the list of peers the lifecycle knows about
```
Jakeeyturner, line 178: Do we need to do this before connecting?
Do we call this code because 'lifecycle' doesn't maintain a list of peers? Is this what they do in the sdk?

cazfletch: no we do it after connecting so that the gateway has set everything up. We need it because we need to tell lifecycle about all the peers so we can use it to later get all the committed smart contracts and all the channels.

PR #2198, opened 2020-04-17
Update environment view channels

cazfletch:


packages/blockchain-extension/extension/commands/installCommand.ts
```diff
@@ -24,7 +22,7 @@ import { FabricEnvironmentRegistryEntry, IFabricEnvironmentConnection, LogType,
 import { Reporter } from '../util/Reporter';
 import { FabricEnvironmentManager } from '../fabric/environments/FabricEnvironmentManager';
 
-export async function installSmartContract(treeItem?: BlockchainTreeItem, peerNames?: Set<string>, chosenPackage?: PackageRegistryEntry): Promise<PackageRegistryEntry | undefined> {
+export async function installSmartContract(peerNames?: Set<string>, chosenPackage?: PackageRegistryEntry): Promise<PackageRegistryEntry | undefined> {
```
Jakeeyturner, line 25: Either peerNames is required or we need to check whether peerNames is undefined/has length 0 and do something

cazfletch: it is correct how it is, if you don't have any peers then the quick pick will show them all

Issue #1778, opened 2019-12-20
As a user i can create a smart contract for fabric 2.0


cazfletch:
When a user clicks create smart contract then it will create one that will target a 2.0 fabric

cazfletch:
done so closing

PR #2215, opened 2020-04-22
Make Generated tests work

cazfletch:
Updated generated tests
Updated generator version to create v2 contracts

contributes to #1778 and #1250


packages/blockchain-extension/extension/commands/testSmartContractCommand.ts
```diff
@@ -535,7 +535,7 @@ async function updateBuildFile(buildFile: string, state: any): Promise<void> {
             'fabric-gateway-java': {
                 groupId: 'org.hyperledger.fabric',
                 artifactId: 'fabric-gateway-java',
-                version: '1.4.2',
+                version: '2.1.0',
```
Jakeeyturner, line 538: It wasn't previously, but should `version: FABRIC_NETWORK_VERSION`?

packages/blockchain-extension/templates/jsTestSmartContractTemplate.ejs
```diff
@@ -29,7 +29,7 @@ const path = require('path');
 <%_ }-%>
 
 <%_ if (contractName !== '') { _%>
-describe('<%=contractName%>-<%=chaincodeLabel%>' , () => {
+describe('<%=contractName%>-<%=chaincodeLabel%>' , async () => {
```
Jakeeyturner, line 32: Why does this need to be `async`?

packages/blockchain-extension/extension/commands/testSmartContractCommand.ts
cazfletch: good point but i think we need separate variables for java and node

packages/blockchain-extension/templates/jsTestSmartContractTemplate.ejs
cazfletch: it doesn't lol

PR #2225, opened 2020-04-23
Codesever version, Java Extension Pack fix

ODOWDAIBM:
- Updated code server version to 3.1.1
- Added wget on to base image
- Java Extension Pack not installing correctly, so just used essential extensions


ODOWDAIBM:
Opps, too early.

Issue #1911, opened 2020-02-12
Fabric 2: Environment View


cazfletch:
Epic for stories needed for the environment view

![image](https://user-images.githubusercontent.com/23044145/74350266-65c03e00-4dad-11ea-902f-179304c4848e.png)

Note nodes and organisations should be two different lists

cazfletch:
done so closing

PR #2235, opened 2020-04-27
Add deploy command

cazfletch:
Added deploy command to do all commands in one step

contributes to #1911


packages/blockchain-extension/extension/commands/deployCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+
+            if (requireCommit) {
+                progress.report({message: `Committing Smart Contract`});
+                await vscode.commands.executeCommand(ExtensionCommands.COMMIT_SMART_CONTRACT, ordererName, channelName, peerNames, smartContractName, smartContractVersion, sequenceNumber);
+            }
+
+            outputAdapter.log(LogType.SUCCESS, 'Successfully deployed smart contract');
+        });
+
+    } catch (error) {
```
Jakeeyturner, line 58: Is it worth us being more specific about where it fails? Just thinking from a debugging perspective, it might be nice if it was to say `"Failed deploying smart contract at 'approving' stage: ${error.messsage}"` as an example.

Jakeeyturner: Although I guess the command which throws the error has a message like this already - nevermind!

packages/blockchain-extension/extension/commands/deployCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+    try {
+        await vscode.window.withProgress({
+            location: vscode.ProgressLocation.Notification,
+            title: 'IBM Blockchain Platform Extension',
+            cancellable: false
+        }, async (progress: vscode.Progress<{ message: string }>) => {
+
+            let packageId: string;
+
+            await vscode.commands.executeCommand(ExtensionCommands.CONNECT_TO_ENVIRONMENT, fabricEnvironmentRegistryEntry);
```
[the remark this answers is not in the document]
cazfletch: no we don't need to as we know we need to connect to this environment to do the deploy. We should already be connected anyway from opening the deploy view

packages/blockchain-extension/extension/commands/deployCommand.ts
cazfletch: you will get other pop ups from command it failed on

Issue #2240, opened 2020-04-28
Fabric 2: Deploy with an endorsement policy


cazfletch:
Update the deploy command to allow a user to pass in an endorsement policy

cazfletch:
done so closing

PR #2256, opened 2020-05-04
Add policy to fabric-admin

cazfletch:
Added new policy parser to use same syntax as cli
Updated Lifecycle channel to use Policy
Added cucumber test

contributes to #2240


packages/blockchain-fabric-admin/src/Policy.ts
```diff
@@ -0,0 +1,151 @@
...
+        return envelope;
+    }
+
+    private parsePolicyString(policy: string): any {
+        return parse(policy);
+    }
+
+    private findIdentityIndex(identity: Identity): number {
+        return this.identityArray.findIndex((otherIdentity) => {
+            return otherIdentity.mspid === identity.mspid && otherIdentity.role === identity.role;
```
Jakeeyturner, line 80: Is there not a name we can check here?

PR #2260, opened 2020-05-04
fixed mdToPdf script

au-shah:


cazfletch:
i think this is ok, will approve after the release

PR #2323, opened 2020-05-20
Added exportAppData command to allow users to export client app data

au-shah:
closes #2220 
Signed-off-by: Akshat Shah <akshatshah@akshats-mbp.hursley.uk.ibm.com>

PR #2426, opened 2020-06-03
Added exportAppData command to allow users to export client app data (bp#2323)

au-shah:
bp #2323 


packages/blockchain-wallet/src/FabricWallet.ts
```diff
@@ -52,6 +52,41 @@ export class FabricWallet implements IFabricWallet {
         return  this.wallet.list();
     }
 
+    public async getIDs(identities: string[]): Promise<FabricIdentity[]> {
+        const allIDs: FabricIdentity[] = [];
```
cazfletch, line 56: this won't work with v2, see the `getIdentities` function for what it should look like

PR #2537, opened 2020-07-20
update ibm cloud group name and behaviour

### LINKED ISSUE #2521: Environment group name "IBM Cloud" is potentially misleading
Currently the "IBM Cloud" environment group can have two types of environment in it: ops tools environments and SaaS environments. But only the SaaS environments are guaranteed to run on IBM Cloud, whereas ops tools environments could potentially be running elsewhere. 

This also causes a problem for the `+ Log in to IBM Cloud` tree item, as the logic that determines whether or not it shows up is impacted by environments that might not even be running on IBM Cloud.

We should consider renaming the IBM Cloud group to something else, or moving the ops tools environments to the "other networks" group or a new group of its own.

erin-hughes:
Closes #2521 


erin-hughes:
We should wait to merge this until the tutorials have been updated to match the changes!

PR #2546, opened 2020-07-21
Fallback GitHub info retrieval (close #2543)

### LINKED ISSUE #2543: Extension will fail to activate if github is down or unavailable
The following error is reported
```
[7/1/2020 1:50:50 PM] [ERROR] Failed to activate extension: Error: Could not get electron version, info.filter is not a function
```
This method `private async getLocalRebuildInfo` has a call to github
```
                const response: any = await Axios.get('https://raw.githubusercontent.com/electron/releases/master/lite.json');
                let info: any[] = response.data;
```
but it looks like it always assumes the call will work.

If github is down or unavailable then maybe the extension could output a meaningful warning and just drop through to rebuilding grpc rather than just failing

Jakeeyturner:

### DIFF
--- a/packages/blockchain-extension/package.json
+++ b/packages/blockchain-extension/package.json
@@ -1407,7 +1407,7 @@
     "scripts": {
         "vscode:prepublish": "npm run compile && npm run localize",
         "package": "node ./node_modules/vsce/out/vsce package",
-        "compile": "rimraf build && npm run build && npm run lint && npm run licchk",
+        "compile": "rimraf build && npm run build && node ./scripts/copyFiles.js && npm run lint && npm run licchk",
         "watch": "rimraf build && tsc -watch",
         "pretest": "npm run updatePackageJSON",
         "test": "node ./build/test/runTests.js",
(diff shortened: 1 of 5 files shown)


packages/blockchain-extension/package.json
```diff
@@ -1599,4 +1599,4 @@
             }
         }
     }
-}
+}
```
erin-hughes, line 1602: Should you have a new line at the end of this file?

packages/blockchain-extension/scripts/copyFiles.js
```diff
@@ -0,0 +1,7 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+const src = path.join(__dirname, '..', 'fallback-build-info.json');
+const dest = path.join(__dirname, '..', 'build', 'fallback-build-info.json');
+
+fs.copySync(src, dest)
```
erin-hughes, line 7: Missing a semi-colon and a new line at the end of the file

PR #2548, opened 2020-07-22
Error popup updated

### LINKED ISSUE #1964: Docker logs shown when transaction fails
## Description

For the Local Fabric, when you submit a transaction which fails, you are switched to the docker logs output channel.
This isn't always the easiest to read and it might be difficult to find out what the error is. Switching to the Blockchain output channel should give a clear reason as to why the transaction failed.

So we should remove the part in our code where we switch to the docker output channel.
## Steps to Reproduce

1.
2.
3.

## Existing issues

## Your Environment

* Visual Studio Code version:
* Link to your project:

### LINKED ISSUE #2172: As a user, I want to know why my environment failed to start more easily
## Description

As a user, it can be frustrating when a network fails to start and the popup message only contains something such as `Failed to start 1 Org Local Fabric: Failed to execute command "/bin/sh" with  arguments "generate.sh" return code 125`.
This isn't particularly consumable, and unless I know to look in the Output section, I won't have any clue as to what this error means.

I suggest we change our popup error message to show a button which says something such as "More Details". By clicking that button, it could bring the Output channel into focus - making it easier to see the underlying issue. Additionally, I suggest we change the message in the popup to be something like: 
`Failed to start 1 Org Local Fabric: Failed to execute command "/bin/sh" with  arguments "generate.sh" return code 125. Check the Output for more details`.

 

## Steps to Reproduce

1.
2.
3.

## Existing issues

## Your Environment

* Visual Studio Code version:
* Link to your project:

kemi04:
Closes #2172
Closes #1964

### DIFF
--- a/packages/blockchain-extension/extension/commands/UserInputUtil.ts
+++ b/packages/blockchain-extension/extension/commands/UserInputUtil.ts
@@ -56,6 +56,7 @@ export class UserInputUtil {
     static readonly NO: string = 'No';
     static readonly DEFAULT: string = 'Default';
     static readonly CUSTOM: string = 'Custom';
+    static readonly MORE_DETAILS: string = 'More Details'
 
     static readonly DEFAULT_SC_EP: string = `${UserInputUtil.DEFAULT} (single endorser, any org)`;
 
@@ -1081,6 +1082,17 @@ export class UserInputUtil {
             await vscode.commands.executeCommand('vscode.open', uri);
         }
     }
+
+    public static async failedNetworkStart(popupMsg: string, logMsg: string): Promise<void> {
+
+        const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+        outputAdapter.log(LogType.ERROR, undefined, logMsg);
+        const response: string = await vscode.window.showErrorMessage(popupMsg, UserInputUtil.MORE_DETAILS);
+        if (response === UserInputUtil.MORE_DETAILS) {
+            outputAdapter.show();
+        }
+     }
+
 
     /**
      * Method to determine if there are multiple smart contracts within the active workspace.
--- a/packages/blockchain-extension/extension/commands/restartFabricRuntime.ts
+++ b/packages/blockchain-extension/extension/commands/restartFabricRuntime.ts
@@ -77,7 +77,7 @@ export async function restartFabricRuntime(runtimeTreeItem?: RuntimeTreeItem): P
         try {
             await runtime.restart(outputAdapter);
         } catch (error) {
-            outputAdapter.log(LogType.ERROR, `Failed to restart ${runtime.getName()}: ${error.message}`, `Failed to restart ${runtime.getName()}: ${error.toString()}`);
+            await UserInputUtil.failedNetworkStart(`Failed to restart ${runtime.getName()}: ${error.message}`, `Failed to restart ${runtime.getName()}: ${error.toString()}`);
         }
     });
 }
--- a/packages/blockchain-extension/extension/commands/startFabricRuntime.ts
+++ b/packages/blockchain-extension/extension/commands/startFabricRuntime.ts
@@ -71,7 +71,7 @@ export async function startFabricRuntime(registryEntry?: RuntimeTreeItem | Fabri
             }
             await runtime.start(outputAdapter);
         } catch (error) {
-            outputAdapter.log(LogType.ERROR, `Failed to start ${runtime.getName()}: ${error.message}`, `Failed to start ${runtime.getName()}: ${error.toString()}`);
+            await UserInputUtil.failedNetworkStart(`Failed to start ${runtime.getName()}: ${error.message}`, `Failed to start ${runtime.getName()}: ${error.toString()}`);
         }
 
         await TimerUtil.sleep(1000);
--- a/packages/blockchain-extension/test/commands/UserInputUtil.test.ts
+++ b/packages/blockchain-extension/test/commands/UserInputUtil.test.ts
@@ -2715,6 +2715,26 @@ describe('UserInputUtil', () => {
         });
     });
 
+    describe('failedNetworkStart', () => {
+
+         it('should display a message passed and the More Details button', async () => {
+            const showErrorMessageStub: sinon.SinonStub = mySandBox.stub(vscode.window, 'showErrorMessage').resolves();
+            const showStub: sinon.SinonStub = mySandBox.stub(VSCodeBlockchainOutputAdapter.instance(), 'show');
+            await UserInputUtil.failedNetworkStart('some error', 'some error message');
+            showErrorMessageStub.should.have.been.calledOnceWithExactly('some error', UserInputUtil.MORE_DETAILS);
+            showStub.should.not.have.been.called;
+         });
+
+         it('should display the Output channel if selected', async () => {
+            const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+            const showErrorMessageStub: sinon.SinonStub = mySandBox.stub(vscode.window, 'showErrorMessage').resolves(UserInputUtil.MORE_DETAILS);
+            const showOutputChannelStub: sinon.SinonStub = mySandBox.stub(outputAdapter, 'show').resolves();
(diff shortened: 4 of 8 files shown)


packages/blockchain-extension/extension/commands/UserInputUtil.ts
```diff
@@ -1082,6 +1082,17 @@ export class UserInputUtil {
         }
     }
 
+    public static async failedNetworkStart(popupMsg: string, logMsg: string, outputAdapter: VSCodeBlockchainOutputAdapter): Promise<void> {
```
Jakeeyturner, line 1085: No need to pass outputAdapter through the function - you can just call:
`const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();`

packages/blockchain-extension/extension/commands/UserInputUtil.ts
```diff
@@ -1082,6 +1082,17 @@ export class UserInputUtil {
...
+
+         const moreInfo: string = 'More Details';
+         outputAdapter.appendLine(LogType.ERROR, logMsg, false);
```
Jakeeyturner, line 1088: You could use `outputAdapter.log` here.
```
VSCodeOutputAdapter.log(type: LogType, popupMessage: string, outputMessage?: string, stackTrace?: string, skipNextLine?: boolean): void
```
So if you just wanted to log something to the output box you could do:
`outputAdapter.log(LogType.ERROR, undefined, logMsg)`

packages/blockchain-extension/extension/commands/restartFabricRuntime.ts
```diff
@@ -77,7 +77,7 @@ export async function restartFabricRuntime(runtimeTreeItem?: RuntimeTreeItem): P
         try {
             await runtime.restart(outputAdapter);
         } catch (error) {
-            outputAdapter.log(LogType.ERROR, `Failed to restart ${runtime.getName()}: ${error.message}`, `Failed to restart ${runtime.getName()}: ${error.toString()}`);
+            await UserInputUtil.failedNetworkStart(`Failed to restart ${runtime.getName()}: ${error.message}`, `Failed to restart ${runtime.getName()}: ${error.toString()}`, outputAdapter);
```
Jakeeyturner, line 80: No need to pass the outputAdapter through here if you make the change to `failedNetworkStart`.

packages/blockchain-extension/extension/commands/startFabricRuntime.ts
```diff
@@ -71,7 +71,7 @@ export async function startFabricRuntime(registryEntry?: RuntimeTreeItem | Fabri
             }
             await runtime.start(outputAdapter);
         } catch (error) {
-            outputAdapter.log(LogType.ERROR, `Failed to start ${runtime.getName()}: ${error.message}`, `Failed to start ${runtime.getName()}: ${error.toString()}`);
+            await UserInputUtil.failedNetworkStart(`Failed to start ${runtime.getName()}: ${error.message}`, `Failed to start ${runtime.getName()}: ${error.toString()}` , outputAdapter);
```
Jakeeyturner, line 74: No need to pass the outputAdapter through here if you make the change to failedNetworkStart.

packages/blockchain-extension/test/commands/UserInputUtil.test.ts
```diff
@@ -2715,6 +2715,25 @@ describe('UserInputUtil', () => {
         });
     });
 
+    describe('failedNetworkStart', () => {
+
+         it('should display a message passed and the More Details button', async () => {
+            const showErrorMessageStub: sinon.SinonStub = mySandBox.stub(vscode.window, 'showErrorMessage').resolves();
+            const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+            await UserInputUtil.failedNetworkStart('some error', 'some error message', outputAdapter);
+            showErrorMessageStub.should.have.been.calledOnceWithExactly('some error', 'More Details');
```
Jakeeyturner, line 2724: As the `showErrorMessageStub` stub is just resolving (which means that the user has cancelled / closed the popup), this means that `outputAdapter.show();` hasn't been called.

So you could do something like:
```
showStub = mySandBox.stub(VSCodeBlockchainOutputAdapter.instance(), 'show');

await UserInputUtil.failedNetworkStart('some error', 'some error message');

showStub.should.not.have.been.called;
```

We should be as strict as possible in our tests!

packages/blockchain-extension/extension/commands/UserInputUtil.ts
```diff
@@ -1082,6 +1082,17 @@ export class UserInputUtil {
         }
     }
 
+    public static async failedNetworkStart(popupMsg: string, logMsg: string, outputAdapter: VSCodeBlockchainOutputAdapter): Promise<void> {
+
+         const moreInfo: string = 'More Details';
```
Jakeeyturner, line 1087: Might be worth making this a static variable at the top of `UserInputUtil.ts` as I expect we might use this multiple times in future.

E.g
`static readonly MORE_DETAILS: string = 'More Details'`

This means that in our tests we can use this variable as well - so if the text ever changes, the tests will be valid as well.

packages/blockchain-extension/test/commands/UserInputUtil.test.ts
```diff
@@ -2715,6 +2715,26 @@ describe('UserInputUtil', () => {
...
+            const showStub: sinon.SinonStub = mySandBox.stub(VSCodeBlockchainOutputAdapter.instance(), 'show');
+            await UserInputUtil.failedNetworkStart('some error', 'some error message');
+            showErrorMessageStub.should.have.been.calledOnceWithExactly('some error', UserInputUtil.MORE_DETAILS);
+            showStub.should.not.have.been.called;
+         });
+
+         it('should display the Output channel if selected', async () => {
+            const outputAdapter: VSCodeBlockchainOutputAdapter = VSCodeBlockchainOutputAdapter.instance();
+            const showErrorMessageStub: sinon.SinonStub = mySandBox.stub(vscode.window, 'showErrorMessage').resolves(UserInputUtil.MORE_DETAILS);
+            const showOutputChannelStub: sinon.SinonStub = mySandBox.stub(outputAdapter, 'show').resolves();
```
lquintai, line 2731: Because we are only using **outputAdapter** to create a stub we don't need to declare it.
We can do the same as what you've done in the previous tests and create the stub directly:

`const showStub: sinon.SinonStub = mySandBox.stub(VSCodeBlockchainOutputAdapter.instance(), 'show').resolves();`

packages/blockchain-extension/test/commands/restartFabricRuntime.test.ts
```diff
@@ -238,9 +240,8 @@ describe('restartFabricRuntime', () => {
 
         executeCommandSpy.should.not.have.been.calledWith(ExtensionCommands.DISCONNECT_GATEWAY);
         executeCommandSpy.should.not.have.been.calledWith(ExtensionCommands.DISCONNECT_ENVIRONMENT);
-
-        logSpy.getCall(0).should.have.been.calledWithExactly(LogType.INFO, undefined, 'restartFabricRuntime');
-        logSpy.getCall(1).should.have.been.calledWithExactly(LogType.ERROR, `Failed to restart ${environment.getName()}: ${error.message}`, `Failed to restart ${environment.getName()}: ${error.toString()}`);
+        popupStub.should.have.been.calledWithExactly(`Failed to restart ${environment.getName()}: ${error.message}`,
+                    `Failed to restart ${environment.getName()}: ${error.toString()}`);
```
lquintai, line 244: This is more of a test structure comment. I've noticed in our team we usually do the following:
 - any prep/create stubs, etc
 - run the command we are interested in testing
 - check the stubs we created for the required functions
 - check the stubs for running/executing other commands
 - check the spies.

Although not necessarily every single tests will be written in this way, I find that most are.
More importantly though, if the other tests in that file follow a structure like this, if you create your tests or add info to existing tests using the same structure, it will make them much easier to read and check in the future, because we'll 'know' where to look for the info.

So the couple of places where you check your popupStub after the spies, they should be in fact after executing the command of interest, and before checking other command executions and spies.

In the tests where you create spies, leaving them at the bottom of the test like you did is the best option.

PR #2574, opened 2020-07-30
Add wallet identity updated

### LINKED ISSUE #755: Improve usability of add connection with enroll ID and secret
The add connection command with an enroll ID and secret asks the following questions:

- What is the identity name?
- What is the MSP ID?
- What is the enrolment ID?
- What is the enrolment secret?

Identity name in a lot of cases may just be the same as the enrolment ID, for example "admin". In which case I end up answering: admin, Org1MSP, admin, adminpw

We could also default the identity name when a certificate is provided, by extracting the common name (CN) from the certificate itself.

@EdMoffatt had some ideas around this

### LINKED ISSUE #1639: Command to remove extension directory

kemi04:
Closes #755
Closes #1639

### DIFF
--- a/packages/blockchain-extension/cucumber/helpers/walletAndIdentityHelper.ts
+++ b/packages/blockchain-extension/cucumber/helpers/walletAndIdentityHelper.ts
@@ -23,6 +23,7 @@ import { UserInputUtil } from '../../extension/commands/UserInputUtil';
 import { UserInputUtilHelper } from './userInputUtilHelper';
 import { ExtensionCommands } from '../../ExtensionCommands';
 import { IFabricWallet, IFabricWalletGenerator, FabricIdentity, FabricWalletRegistry, FabricWalletRegistryEntry, FabricWalletGeneratorFactory, FabricGatewayRegistryEntry } from 'ibm-blockchain-platform-common';
+import * as fs from 'fs-extra';
 chai.use(sinonChai);
 chai.use(chaiAsPromised);
 
@@ -123,19 +124,29 @@ export class WalletAndIdentityHelper {
     }
 
     private setIdentityStubs(method: string, identityName: string, mspid: string): void {
-        this.userInputUtilHelper.inputBoxStub.withArgs('Provide a name for the identity').resolves(identityName);
         this.userInputUtilHelper.inputBoxStub.withArgs('Enter MSPID').resolves(mspid);
 
         if (method === 'certs') {
             this.userInputUtilHelper.showAddIdentityMethodStub.resolves(UserInputUtil.ADD_CERT_KEY_OPTION);
+            this.userInputUtilHelper.inputBoxStub.withArgs('Provide a name for the identity').resolves(identityName);
             this.userInputUtilHelper.showGetCertKeyStub.resolves({ certificatePath: WalletAndIdentityHelper.certPath, privateKeyPath: WalletAndIdentityHelper.keyPath });
         } else if (method === 'JSON file') {
             const jsonPath: string = process.env.OPSTOOLS_FABRIC ? path.join(process.env.JSON_DIR, `${identityName}.json`) : WalletAndIdentityHelper.jsonFilePath;
             this.userInputUtilHelper.showAddIdentityMethodStub.resolves(UserInputUtil.ADD_JSON_ID_OPTION);
+            fs.readFile(jsonPath, 'utf8', function readFileCallback(err: any, data: string): void {
+                if (!err) {
+                    const obj: any = JSON.parse(data);
+                    obj.name = identityName;
+                    const json: any = JSON.stringify(obj);
+                    fs.writeFile(jsonPath, json, 'utf8', () => {
+                        return;
+                    });
+            }});
             this.userInputUtilHelper.browseStub.resolves(vscode.Uri.file(jsonPath));
         } else {
             // use enroll id and secret
             this.userInputUtilHelper.showAddIdentityMethodStub.resolves(UserInputUtil.ADD_ID_SECRET_OPTION);
+            this.userInputUtilHelper.inputBoxStub.withArgs('Provide a name for the identity').resolves(identityName);
             const gatewayRegistryEntry: FabricGatewayRegistryEntry = new FabricGatewayRegistryEntry();
             gatewayRegistryEntry.name = 'myGateway';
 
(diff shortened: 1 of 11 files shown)


packages/blockchain-extension/cucumber/helpers/walletAndIdentityHelper.ts
```diff
@@ -23,6 +23,7 @@ import { UserInputUtil } from '../../extension/commands/UserInputUtil';
 import { UserInputUtilHelper } from './userInputUtilHelper';
 import { ExtensionCommands } from '../../ExtensionCommands';
 import { IFabricWallet, IFabricWalletGenerator, FabricIdentity, FabricWalletRegistry, FabricWalletRegistryEntry, FabricWalletGeneratorFactory, FabricGatewayRegistryEntry } from 'ibm-blockchain-platform-common';
+import * as fs from 'fs';
```
Jakeeyturner, line 26: `import * as fs from 'fs-extra`

packages/blockchain-extension/extension/commands/addWalletIdentityCommand.ts
```diff
@@ -27,6 +27,19 @@ import { ManagedAnsibleEnvironment } from '../fabric/environments/ManagedAnsible
 import { EnvironmentFactory } from '../fabric/environments/EnvironmentFactory';
 import { LocalEnvironment } from '../fabric/environments/LocalEnvironment';
 
+export async function checkIdentityName(identityName: string, identityExists: boolean): Promise<boolean>
```
Jakeeyturner, line 30: No need to export it - we can just treat it like a 'private' function.
In our tests the function will just get called as normal when we're testing the `addWalletIdentityCommand` (no need to test `checkIdentityName` isolated).

packages/blockchain-extension/extension/commands/addWalletIdentityCommand.ts
```diff
@@ -27,6 +27,20 @@ import { ManagedAnsibleEnvironment } from '../fabric/environments/ManagedAnsible
...
+        return false;
+    }
+    
+    if (identityExists) {
+        VSCodeBlockchainOutputAdapter.instance().log(LogType.ERROR, `An identity called ${identityName} already exists`);
+        return false;
+    }
+    return true;
+}
+
```
lquintai, line 43: As is, this function doesn't need to be async, since everything it does currently happens synchronously.

There are 2 options:
1. Remove the async keyword and the 'await' from the places where it is called;
2. Instead of taking 'identityExists' as an input argument, it can instead take in 'wallet'. This means we would evaluate if the identity exists inside this function, and therefore not require `const identityExists: boolean = await wallet.exists(identityName);` to be called in 3 different locations. It would also mean we won't need to check it until we are sure the identity actually exists, failing sooner if it doesn't. It would also make this function async because 'wallet.exists' is async.

Finally, this function should be at the end of the file, since it is not the most important one. The function `addWalletIdentity` is the one that gives this file it's name, so it should be the first thing we read when we look at this file.

packages/blockchain-extension/tutorials/developer-tutorials/add-environment.md
```diff
@@ -48,37 +48,37 @@ The identities that were exported need to be imported into VS Code. First we wil
...
+3. You will then be asked to provide the MSPID use `org1msp` if you followed the tutorial.
 
-4. You will then be asked to provide the MSPID use `org1msp` if you followed the tutorial.
+4. You will then be asked for the method for adding an identity choose `Provide JSON identity file from IBM Blockchain Platform`.
 
-5. You will then be asked for the method for adding an identity chose `Provide JSON identity file from IBM Blockchain Platform`.
+5. Browse to the location of the previously exported `Org1 Admin` file and select it.
 
-6. Browse to the location of the previously exported `Org1 Admin` file and select it.
+6. Alternatively, If the JSON identity file does not contain the name property, you will be asked to enter a name for the identity.
```
lquintai, line 57: We're saying `Alternatively`, but not what this is an alternative to. Perhaps something like this:
6. If the JSON identity file does not contain the name property, you will be asked to enter a name for the identity, otherwise this step is skipped.

packages/blockchain-extension/tutorials/developer-tutorials/add-environment.md
```diff
@@ -48,37 +48,37 @@ The identities that were exported need to be imported into VS Code. First we wil
...
+2. You will then be asked to provide an MSPID use `osmsp`.
 
-3. You will then be asked to provide an MSPID use `osmsp`.
+3. You will then be asked for the method for adding an identity choose `Provide JSON identity file from IBM Blockchain Plaform`.
 
-4. You will then be asked for the method for adding an identity chose `Provide JSON identity file from IBM Blockchain Plaform`.
+4. Browse to the location of the previously exported `Orderer Admin` file and select it.
 
-6. Browse to the location of the previously exported `Orderer Admin` file and select it.
+5. Alternatively, If the JSON identity file does not contain the name property, you will be asked to enter a name for the identity.
```
lquintai, line 69: as above

packages/blockchain-extension/tutorials/developer-tutorials/add-environment.md
```diff
@@ -48,37 +48,37 @@ The identities that were exported need to be imported into VS Code. First we wil
...
+2. You will then be asked to provide an MSPID. Use `org1msp` here.
 
-3. You will then be asked to provide an MSPID. Use `org1msp` here.
+3. You will then be asked for the method for adding an identity, select `Provide JSON identity file from IBM Blockchain Plaform`.
 
-4. You will then be asked for the method for adding an identity, select `Provide JSON identity file from IBM Blockchain Plaform`.
+4. Browse to the location of the previously exported `Org1 CA Admin` file and select it.
 
-6. Browse to the location of the previously exported `Org1 CA Admin` file and select it.
+5. Alternatively, If the JSON identity file does not contain the name property, you will be asked to enter a name for the identity.
```
lquintai, line 81: as above

packages/blockchain-extension/cucumber/helpers/walletAndIdentityHelper.ts
```diff
@@ -123,19 +124,29 @@ export class WalletAndIdentityHelper {
...
             this.userInputUtilHelper.showAddIdentityMethodStub.resolves(UserInputUtil.ADD_JSON_ID_OPTION);
+            fs.readFile(jsonPath, 'utf8', function readFileCallback(err: any, data: string): void {
+                if (!err) {
+                    const obj: any = JSON.parse(data);
+                    obj.name = identityName;
+                    const json: any = JSON.stringify(obj);
+                    fs.writeFile(jsonPath, json, 'utf8', () => {
+                        return;
+                    });
+            }});
```
lquintai, line 144: This worries me, because we shouldn't need to read/modify/write our identities. If the read or the write fail the test will fail for an unrelated reason.

Instead we should make sure that our tests are correct and relevant, which in this case *might* mean renaming the identities we are expecting to create.

I'm happy to discuss this further via DM.

PR #2584, opened 2020-08-03
[v1] - fix login prompts if logged in with no account

### LINKED ISSUE #2583: [BUG] user unable to access environments panel if loged in to IBM cloud but account not selected
How to recreate:
1. login to IBM cloud (any mechanism) and do NOT select account
2. refresh window or restart extension - user will be prompted to select account. If none selected, view hungs and user is unable to access environments panel.

Solution:
If the user is logged in but there is no account selected, instead of immediately prompt the user to select account we display a tree item that reads `+ Select IBM Cloud account`.
By clicking on this tree item the user will be asked to select an account from a list created by the cloud account extension.

Since the user is already logged in, they'll see `IBM Cloud: logged in` on the bottom left corner as expected.

lquintai:
Closes #2583  


Jakeeyturner (changes requested):
When I login (using SSO), it asks me to select an account from my list of two linked accounts.
When I cancel out, I get a success message `Logged into IBM Cloud as undefined`.

I didn't proceed any further from this as I suspect it'll error.

For context I'm also using v1.0.2 of the IBM Cloud extension.

lquintai:
Indeed, I found the same problem so I created a PR to the cloud extension to fix it, and it's now included in v1.0.3.
Sadly, I don't think we can specify the exact version we want to install from our extension in the extensionPack section of package.json, I think we just always get the latest.

Having said that, I'm not sure if we will get the latest at upgrade, which would solve the problem, so investigating.

From the extension manifest docs: 
`extensionPack | array |  An array with the ids of extensions bundled with this extension.  These other extensions will be installed when the primary extension is  installed. The id of an extension is always ${publisher}.${name}. For example: vscode.csharp.`

Jakeeyturner (approved):

PR #2585, opened 2020-08-05
[v2] - Connect to environment/gateway if at least one of the channels is v2 enabled.

### LINKED ISSUE #2540: Handle connecting to network with 1.4 & 2.0 channels (v2)
## Description

In a users network there might be numerous channels, each with different capabilities enabled.
At the moment, if any channels aren't V2, then we throw an error and prevent the user from connecting.

As people start to migrate from Fabric 1.4 to 2, it's likely that they'll have a mix of both - so we should handle this behaviour.

One suggestion is that instead of erroring out, we could warn the user that they have a mix of channels but only show/allow interaction with V2 channels.
Another suggestion is to show an icon next to the V1.4 channels.

We also need to make sure we cover the case where all channels are v1.

Will need some design input on this! @Charlotte8699 @EdMoffatt 

## Decision
Display only v2 channels, and alert the user that v1 channels were present and are not going to be displayed.

This is needed to allow users with mixed channels, however the behaviour will change once v1->v2 merge is one.

lquintai:
Implementation:

When there are no v2 enabled channels:
 - the user will be unable to connect to the environment or gateway
 - an error will be displayed.

When there are a mix of v1 and v2 channels:
 - the user will be able to connect to the environment or gateway
 - a warning will be displayed to the user saying v1 channels were detected, and the names of those channels
 - only the v2 channels will be visible in the tree

Closes #2540 


packages/blockchain-environment-v1/src/FabricEnvironmentConnection.ts
```diff
@@ -198,7 +198,11 @@ export class FabricEnvironmentConnection implements IFabricEnvironmentConnection
                     const peer: LifecyclePeer = await this.getPeer(peerName);
                     const capabilities: string[] = await peer.getChannelCapabilities(channelName);
                     if (!capabilities.includes('V2_0')) {
-                        throw new Error(`channel '${channelName}' does not have V2_0 capabilities enabled.`);
+                        if (!v1channels.includes(channelName)) {
+                            v1channels.push(channelName);
+                        }
+                        continue;
+                        // throw new Error(`channel '${channelName}' does not have V2_0 capabilities enabled.`);
```
Jakeeyturner, line 205: Comment can be deleted

packages/blockchain-gateway-v1/src/FabricConnection.ts
```diff
@@ -117,7 +119,9 @@ export abstract class FabricConnection {
                     const peer: LifecyclePeer = this.lifecycle.getPeer(peerName, this.gateway.getOptions().wallet, this.gateway.getOptions().identity as string);
                     const capabilities: string[] = await peer.getChannelCapabilities(channelName);
                     if (!capabilities.includes('V2_0')) {
-                        throw new Error(`channel '${channelName}' does not have V2_0 capabilities enabled.`);
+                        v1channels.push(channelName);
+                        continue;
+                        // throw new Error(`channel '${channelName}' does not have V2_0 capabilities enabled.`);
```
Jakeeyturner, line 124: Comment can be deleted

lquintai: Well spotted!

Jakeeyturner (approved):

PR #2588, opened 2020-08-06
Delete directory command added

### LINKED ISSUE #1639: Command to remove extension directory

kemi04:
Closes #1639

### DIFF
--- a/packages/blockchain-extension/extension/commands/deleteExtensionDirectoryCommand.ts
+++ b/packages/blockchain-extension/extension/commands/deleteExtensionDirectoryCommand.ts
@@ -1,0 +1,40 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


packages/blockchain-extension/extension/commands/deleteExtensionDirectoryCommand.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ExtensionCommands } from '../../ExtensionCommands';
+import { VSCodeBlockchainOutputAdapter } from '../logging/VSCodeBlockchainOutputAdapter';
+import { FabricEnvironmentRegistryEntry, LogType, FabricEnvironmentRegistry } from 'ibm-blockchain-platform-common';
+import { SettingConfigurations } from '../../configurations';
+import { UserInputUtil } from './UserInputUtil';
+
+export async function deleteExtensionDirectory(): Promise<void> {
+
+    try {
+        if (!await UserInputUtil.showConfirmationWarningMessage(`This will delete the extension directory. Do you want to continue?`)) {
```
lquintai, line 26: This is correct, and from a development point of view shorter and easy to understand, so I know why you wrote it this way.
However, in our extension we rarely use commands as a condition directly, instead we run the command, assign the return value to a variable, and then say `if (variable)`.

The reason we do this is because it becomes easier to test. When running tests we can see exactly what the return value from that function was, and it makes it easier to spot errors.
For example, the funtion might change and instead of a boolean it now returns a string, or an object. We might not necessarily spot this if the function is used as a condition directly.

In this case, the user canceling (void) or saying no (false) leads to the same outcome, stop/abort executing the delete directory command. This is not always the case though, sometimes if we get false we follow a specific code path, if we get void we abort, for example.

So even thought it is perfectly correct, changing it to store the return value in a variable and use the variable as a condition is more in line with the rest of the extension's code.

PR #2594, opened 2020-08-11
Delete directory command added

### LINKED ISSUE #1639: Command to remove extension directory

kemi04:
Closes #1639

### DIFF
--- a/packages/blockchain-extension/extension/commands/deleteExtensionDirectoryCommand.ts
+++ b/packages/blockchain-extension/extension/commands/deleteExtensionDirectoryCommand.ts
@@ -1,0 +1,43 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


packages/blockchain-extension/extension/commands/deleteExtensionDirectoryCommand.ts
```diff
@@ -0,0 +1,43 @@
...
+        }
+        const environments: FabricEnvironmentRegistryEntry[] = await FabricEnvironmentRegistry.instance().getAll(true);
+        for (const environment of environments) {
+            if (environment.environmentType === EnvironmentType.LOCAL_ENVIRONMENT) {
+                await vscode.commands.executeCommand(ExtensionCommands.TEARDOWN_FABRIC, undefined, true, environment.name);
+                await FabricEnvironmentRegistry.instance().delete(environment.name);
+            }
+        }
+        const extDir: string = vscode.workspace.getConfiguration().get(SettingConfigurations.EXTENSION_DIRECTORY);
+        await fs.remove(extDir);
```
lquintai, line 38: The value returned by vscode.workspace.getConfiguration().get(SettingConfigurations.EXTENSION_DIRECTORY) needs to be transformed to something the filesystem understands. We can use this:

`const resolvedExtDir: string = FileSystemUtil.getDirPath(extDir);`

Sadly _fs.remove_ silently does nothing if the path provided doesn't exist, and _fs.rmdir_ is not recursive on the supported versions of node, so it can't be used on non-empty folders.

So, the way around it would be to check if the path exists, and if it doesn't, thrown an error, which will stop the execution of the command and it should:

```
const exists = await fs.pathExists(resolvedExtDir);
if (!exists) {
     throw new Error('some meaningful error text here');
}
```
And then go ahead with the removal, because if we get here we confirmed the path exists:
`await fs.remove(resolvedExtDir);`

PR #2612, opened 2020-08-20
Exported Connection profile names updated

### LINKED ISSUE #2175: As a user, it would be better to export my connection profile in pascal case
## Description

MGK came across this problem where he exported his connection profile and opened it from within VS Code, which threw an error. This is due to the spaces in the connection profile name.
image (10)
image (11)

As a fix, I suggest we export the file in pascal case, e.g. `1OrgLocalFabricOrg1Connection.json`. This will prevent this error, in addition to making it much to interact with inside a terminal (and lab document), where tab-autocomplete currently does this: `$ cat 1\ Org\ Local\ Fabric\ -\ Org1_connection.json`

Where we export anything else, it might be worth exporting in the same pascal case format!

## Update
After discussing the issue, it was decided not to use the pascal case, but only to remove spaces as it can be confusing if a user uses another naming convention style for his environment and wallet names.

kemi04:
Closes #2175

### DIFF
--- a/packages/blockchain-extension/extension/commands/exportConnectionProfileCommand.ts
+++ b/packages/blockchain-extension/extension/commands/exportConnectionProfileCommand.ts
@@ -26,6 +26,7 @@ import { GatewayTreeItem } from '../explorer/model/GatewayTreeItem';
 import { FabricGatewayHelper } from '../fabric/FabricGatewayHelper';
 import { FabricGatewayConnectionManager } from '../fabric/FabricGatewayConnectionManager';
 import { ConnectionProfileUtil, LogType, FabricGatewayRegistryEntry } from 'ibm-blockchain-platform-common';
+import * as lodash from 'lodash';
 
 export async function exportConnectionProfile(gatewayTreeItem: GatewayTreeItem, isConnected?: boolean): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/blockchain-extension/extension/commands/exportConnectionProfileCommand.ts
```diff
@@ -51,7 +51,8 @@ export async function exportConnectionProfile(gatewayTreeItem: GatewayTreeItem,
     // Ask the user where they want to export it to
     // set the default path to be the first open workspace folder
     let defaultPath: string;
-    const fileName: string = `${gatewayEntry.name}_connection.json`;
+    let fileName: string = `${gatewayEntry.name}_connection.json`;
+    fileName = fileName.replace(/\s/g, '');
```
lquintai, line 55: This removed white spaces, but does not create a upper first camel case as requested.

I created a gateway with the following terrible name: _attempt 1 my one-org local fabric GATEWAY_

The current code will create a file name that is very hard to read: `attempt 1 my one-org local fabric GATEWAY_connection.json`

In upper camel case (where all words start with upper case, followed by lower case, hyphens and spaces are ignored, and numbers are not touched), this would be much more readable: `Attempt1MyOneOrgLocalFabricGatewayConnection.json`

You can achieve this with _upperFirst_ and _camelCase_ from the lodash package for example.

lquintai (approved):
LGTM.

PR #2614, opened 2020-08-20
GOPATH issue updated

### LINKED ISSUE #2596: GOPATH issues when packaging
https://github.com/IBM-Blockchain/blockchain-vscode-extension/blob/13b23135fbedf8589d8e9e5fa2ab329cb5be3900/packages/blockchain-extension/extension/commands/packageSmartContractCommand.ts#L117

`GOPATH` can actually contain a set of directories separated by the system path separator (`:` on Linux/macOS, `;` on Windows). This isn't handled by this code. We need to scan all of the directories in the `GOPATH` environment variable looking to see if the project directory is under `GOPATH`.

In the case that `GOPATH` is not set, we check that `<project directory>/../../src` exists. This assumes one path element and only works for `~/projects/src/sstone1/myproject`. There can be any number of path elements between the `src` directory and the project directory, for example a common Go directory structure is ~/projects/src/github.com/sstone1/myproject`.

kemi04:
Closes #2596

### DIFF
--- a/packages/blockchain-extension/extension/commands/packageSmartContractCommand.ts
+++ b/packages/blockchain-extension/extension/commands/packageSmartContractCommand.ts
@@ -97,6 +97,7 @@ export async function packageSmartContract(workspace?: vscode.WorkspaceFolder, o
         cancellable: false
     }, async (progress: vscode.Progress<{ message: string }>) => {
         progress.report({ message: `Packaging Smart Contract` });
+        let originalGOPATH: string = '';
         try {
 
             // Determine the filename of the new package.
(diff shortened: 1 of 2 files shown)


packages/blockchain-extension/test/commands/packageSmartContractCommand.test.ts
```diff
@@ -39,20 +39,23 @@ describe('packageSmartContract', () => {
...
         let projectDir: string;
         if (language === 'golang') {
             if (createWrongPlace) {
                 folders[2].uri = vscode.Uri.file(wrongGolangPath);
                 projectDir = path.join(testWorkspace, packageName);
-            } else {
+            } else if (srcChild) {
                 projectDir = path.join(testWorkspace, 'src', packageName);
+            } else {
+                projectDir = path.join(testWorkspace, 'src', 'child', packageName);
```
lquintai, line 58: Just a suggestion  to change the folder name between src and packageName to something else. It becomes confusing when later on tests that mention something is not a child to then have child in the name. It's not wrong, just a bit confusing and makes understanding the tests harder.

packages/blockchain-extension/test/commands/packageSmartContractCommand.test.ts
```diff
@@ -940,6 +973,61 @@ describe('packageSmartContract', () => {
...
+            });
+
+            findFilesStub.withArgs(new vscode.RelativePattern(folders[testIndex], '**/*.go'), null, 1).resolves([vscode.Uri.file('chaincode.go')]);
+
+            showInputStub.onFirstCall().resolves('myProject');
+            showInputStub.onSecondCall().resolves('0.0.3');
+            showInputStub.onThirdCall().resolves('myProject');
+            showInputStub.onCall(3).resolves('0.0.3');
+
+            process.env.GOPATH = '/' + path.delimiter + testWorkspace;
```
lquintai, line 993: Similar to the previous suggestion: change '/' to a path, instead of just one character, say /some/path/'. It's not wrong, but forces the reader to look again and pause, so it takes longer to understand the test.

lquintai (changes requested):
Would be useful to know what tests were performed to check the changes, so that I can try to recreate thm.

packages/blockchain-extension/extension/commands/packageSmartContractCommand.ts
```diff
@@ -115,25 +116,34 @@ export async function packageSmartContract(workspace?: vscode.WorkspaceFolder, o
...
+                    const indexSrc: number = contractPath.indexOf(path.sep + 'src' + path.sep);
+                    const srcPath: string = contractPath.substring(0, indexSrc + 4);
+                    const goPaths: string[] = process.env.GOPATH.split(path.delimiter);
+                    let pathsMatch: boolean = false;
+                    if (goPaths.length > 1) {
+                        originalGOPATH = process.env.GOPATH;
+                    }
+                    goPaths.forEach((value: string) => {
+                        if (value === srcPath.substr(0, srcPath.length - 4)) {
+                            process.env.GOPATH = value;
```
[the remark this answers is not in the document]
kemi04: There is a problem with having more than 1 path in process.env.GOPATH : when there are multiple paths, Package.fromDirectory  from fabric-client  node module returns file not found error:
ENOENT: no such file or directory, scandir '/:/Users/kestutisgudynas/Desktop/blockchain-vscode-extension/packages/blockchain-extension/test/data/testWorkspace/src/goProject'
In this case, process.env.GOPATH = path.resolve('/' + path.delimiter + testWorkspace);
However, there is a simple workaround:  we can delete all other GOPATH paths and keep only the needed one.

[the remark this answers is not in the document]
lquintai: Makes sense, thanks!

packages/blockchain-extension/extension/commands/packageSmartContractCommand.ts
```diff
@@ -115,25 +116,37 @@ export async function packageSmartContract(workspace?: vscode.WorkspaceFolder, o
...
+                        throw new Error('The Go smart contract is not a subdirectory of the path specified by the environment variable GOPATH. Please correct the environment variable GOPATH.');
+                    }
+                    const srcPath: string = contractPath.substring(0, indexSrc + 4);
+                    const goPaths: string[] = process.env.GOPATH.split(path.delimiter);
+                    let pathsMatch: boolean = false;
+                    if (goPaths.length > 1) {
+                        originalGOPATH = process.env.GOPATH;
+                    }
+                    goPaths.forEach((value: string) => {
+                        if (value === srcPath.substr(0, srcPath.length - 4)) {
```
lquintai, line 143: This implies that the path stored in value does not end with a directory delimiter: `srcPath.substr(0, srcPath.length - 4)`

That is not necessarily true, so we should replace the last delimiter of value if present before comparing.

packages/blockchain-extension/extension/commands/packageSmartContractCommand.ts
```diff
@@ -115,25 +116,37 @@ export async function packageSmartContract(workspace?: vscode.WorkspaceFolder, o
...
+                        contractPath = path.relative(srcPath, contractPath);
                         process.env.GOPATH = path.join(srcPath, '..');
                     }
                 } else {
-                    // The path is relative to $GOPATH/src for Go smart contracts.
-                    const srcPath: string = path.join(process.env.GOPATH, 'src');
+                    // The path is relative to $GOPATH/src for Go smart contracts.
+                    const indexSrc: number = contractPath.indexOf(path.sep + 'src' + path.sep);
+                    if (indexSrc === -1) {
+                        throw new Error('The Go smart contract is not a subdirectory of the path specified by the environment variable GOPATH. Please correct the environment variable GOPATH.');
```
lquintai, line 134: We are throwing exactly the same error below, ideally we don't want to throw the same error in two different places if possible to avoid.
The code below is already throwing the error under four different conditions, it should be easy to add this one and make it so that we only have the error coming from one location in the code.

lquintai (changes requested):

lquintai (approved):
Manual test:
 - no gopath and contract in nested directories inside src
 - gopath with multiple possible paths, delimited by ':' (linux)

LGTM.

PR #2624, opened 2020-08-27
v1 - display only v1 channels

lquintai:
All node modules from the sdk are v1.4.12-snapshot.4. When trying to use v1.4.11 for other sdk packages, got incompatibility problems with fabric-network:

```
src/FabricConnection.ts(115,27): error TS2741: Property 'getChannelCapabilities' is missing in type 'import("<path>/node_modules/fabric-network/node_modules/fabric-client/types/index").Channel' but required in type 'import("<path>/node_modules/fabric-client/types/index").Channel'.
src/FabricConnection.ts(156,19): error TS2322: Type 'import("<path>/node_modules/fabric-network/node_modules/fabric-client/types/index").Channel' is not assignable to type 'import("<path>/node_modules/fabric-client/types/index").Channel'.
src/FabricConnection.ts(194,15): error TS2719: Type 'Client' is not assignable to type 'Client'. Two different types with this name exist, but they are unrelated.
  Types of property 'newChannel' are incompatible.
    Type '(name: string) => import("<path>/node_modules/fabric-network/node_modules/fabric-client/types/index").Channel' is not assignable to type '(name: string) => import("<path>/node_modules/fabric-client/types/index").Channel'.
      Type 'import("<path>/node_modules/fabric-network/node_modules/fabric-client/types/index").Channel' is not assignable to type 'import("<path>/node_modules/fabric-client/types/index").Channel'.
src/FabricConnection.ts(216,19): error TS2322: Type 'import("<path>/node_modules/fabric-network/node_modules/fabric-client/types/index").Channel' is not assignable to type 'import("<path>/node_modules/fabric-client/types/index").Channel'.


```

### Merge or not to merge

Erin is happy with the code, we should be able to merge when we decide what to do regarding the SDK. 

If we are going to wait (might be ready when Jake comes back!) for the new SDK to be released, there are 2 options:

1. When the new SDK release is out, one of you can add a commit to this PR where you modify the multiple package.json files and run lerna clean+bootstrap+compile on it, and then deliver both commits at the same time 
2. Merge this as is, and then open another PR just with the modified package.json and package-lock.json files

### DIFF
--- a/packages/blockchain-common/src/interfaces/IFabricEnvironmentConnection.ts
+++ b/packages/blockchain-common/src/interfaces/IFabricEnvironmentConnection.ts
@@ -28,7 +28,7 @@ export interface IFabricEnvironmentConnection {
 
     getAllPeerNames(): Array<string>;
 
-    createChannelMap(): Promise<Map<string, Array<string>>>;
+    createChannelMap(): Promise<{channelMap: Map<string, Array<string>>, v2channels: Array<string>}>;
 
     getInstantiatedChaincode(peerNames: Array<string>, channelName: string): Promise<Array<FabricChaincode>>;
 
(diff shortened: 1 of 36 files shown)


erin-hughes (commented):
Tried this out using a network with both v1 and v2 channels, and was only able to see / interact with v2 channels. LGTM

Jakeeyturner (approved):

PR #2646, opened 2020-09-22
v1 - Fix dependencies

### LINKED ISSUE #2633: OpenSSL Requirements for windows may be incorrect
The following is in the readme
```
You will need to install OpenSSL v1.0.2 if using Node 8, or v1.1.1 if using Node 10 from Win32 OpenSSL
Install the normal version, not the version marked as "light"
Install the Win32 version into C:\OpenSSL-Win32 on 32-bit systems
Install the Win64 version into C:\OpenSSL-Win64 on 64-bit systems
```

I think first it would be helpful to state that this is optional. It's only required to debug 1.4 node smart contracts.
Secondly I think that only OpenSSL v1.0.2 can be used because only this version contains the static openssl library required by the ampretia/x509 node module (ie v1.1.1 doesn't), not sure if then that would cause a problem if you are using node 10.
Finally I think we should remove the Win32 description, it doesn't make sense to mention any support for Win32 given it's highly unlikely anything here would actually work in a 32-bit environment.

### LINKED ISSUE #2641: [v1] Update required Node version
## Description

I believe for v1 & v2 we need to update the extension to require `^10.15.3 || ^12.13.1` of Node. This is based off of "engines" section in the fabric-sdk-node's `package.json`.

We should obviously check that these versions work for creating, packaging, deploying a contract, etc.

### Release 1.4
"engines": {
    "node": "^10.15.3 || ^12.15.0",
    "npm": "^6.4.1"
  },

### Release 2.2
  "engines": {
    "node": "^10.15.3 || ^12.13.1",
    "npm": "^6.4.1"
  },

lquintai:
Changed node dependencies from versions 8 and 10 to 10 and 12 according to node sdk engine requirements.
Changed openSSL requirement to be version v1.0.2 regardless of node version, and removed windows 32bit information from the readme and prereqs links.

closes #2641 
closes #2633 

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,7 +455,7 @@ The following dependencies are required on all operating systems:
 - [VS Code version 1.38.0 or greater](https://code.visualstudio.com)
 > VS Code version can be found by running: `code --version`
 
-- [Node v8.x or v10.x and npm v6.x or greater](https://nodejs.org/en/download/)
+- [Node v10 (v10.15.3 or greater) or v12 (v12.15.0 or greater) and npm v6.x or greater](https://nodejs.org/en/download/)
 > Node version can be found by running: `node --version`
 >
 > npm version can be found by running: `npm --version`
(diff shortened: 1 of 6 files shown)


erin-hughes (dismissed):
Confirmed that node 8 no longer satisfies the dependency manager, and that upgrading to a valid version is picked up and the extension activates as expected. LGTM!

erin-hughes (approved):

PR #2675, opened 2020-10-01
Improve performance for prereqs view

james-wallis:
*The diff for this looks huge but it's mainly just moving code around + adding interfaces*
### Summary
Currently the prerequisites page takes some time to load, I've attempted to improve performance by refactoring the current codebase by:
- [x] Using `Promise.all` where possible so that system version checks (Docker, Node etc) can occur at the same time.
- [x] Pass the dependencies through to the PreReqView optionally so that the `getPreReqVersions` functions only runs once when the user starts the extension (used to run twice).
    * The Prereqs page will re-scan the dependencies whenever a user opens it using the command pallet.

Also:
* Moved the Dependencies/RequiredDependencies/OptionalDependencies into their own `interfaces`.
* Moved the defaults into the `Dependencies.ts` file so that it is easier to manage what a Dependency is and what they are. 
* Slightly refactored the `PreReqsView` test to use the Dependency interface and have less setup for each test.
* Modified the test coverage if statement as it wasn't being triggered for me.

### Tests
* Ensured all unit tests still work.
* Checked that the prereqs page still appears the same as before.

### DIFF
--- a/packages/blockchain-extension/extension/dependencies/Dependencies.ts
+++ b/packages/blockchain-extension/extension/dependencies/Dependencies.ts
@@ -13,7 +13,7 @@
 */
 'use strict';
 
-export class Dependencies {
+export class DependencyVersions {
 
     static readonly DOCKER_REQUIRED: string = '>=17.6.2';
     static readonly DOCKER_COMPOSE_REQUIRED: string = '>=1.14.0';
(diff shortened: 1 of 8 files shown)


erin-hughes (approved):
Code LGTM, the prereq page also looks and behaves how I'd expect 👌

PR #2694, opened 2020-10-08
v2 - update node prereqs

### LINKED ISSUE #2693: [v2] Update required Node version
## Description

I believe for v1 & v2 we need to update the extension to require `^10.15.3 || ^12.13.1` of Node. This is based off of "engines" section in the fabric-sdk-node's `package.json`.

We should obviously check that these versions work for creating, packaging, deploying a contract, etc.

### Release 1.4
"engines": {
    "node": "^10.15.3 || ^12.15.0",
    "npm": "^6.4.1"
  },

### Release 2.2
  "engines": {
    "node": "^10.15.3 || ^12.13.1",
    "npm": "^6.4.1"
  },

lquintai:
I believe pre-req state is now persistent so I can't modify the dependency directly with the readable version requirements for node - hence the dependency clone for the pre-req view.

closes #2693 

### DIFF
--- a/packages/blockchain-extension/extension/webview/PreReqView.ts
+++ b/packages/blockchain-extension/extension/webview/PreReqView.ts
@@ -24,7 +24,7 @@ import { SettingConfigurations } from '../configurations';
 import { GlobalState, ExtensionData } from '../util/GlobalState';
 import { VSCodeBlockchainOutputAdapter } from '../logging/VSCodeBlockchainOutputAdapter';
 import { LogType } from 'ibm-blockchain-platform-common';
-import { Dependencies } from '../dependencies/Dependencies';
+import { Dependencies, Dependency, DependencyWithVersion } from '../dependencies/Dependencies';
 
 export class PreReqView extends View {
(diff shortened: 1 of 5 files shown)


james-wallis (approved):
Code looks good. I've tested this on Node version 10 (a valid one) and 10.14.2 (invalid) to confirm that the changes work.

packages/blockchain-extension/extension/webview/PreReqView.ts
```diff
@@ -191,12 +191,21 @@ export class PreReqView extends View {
         let missingPrerequisitesLength: number = 0;
 
         for (const _dependency of Object.keys(dependencies)) {
+            let dependency: any = dependencies[_dependency];
```
james-wallis, line 194: If you want to remove the `any` here it could be changed to `Dependency` which is in `'../dependencies/Dependencies';`

james-wallis (approved):
Still looks good! Thanks for changing `any` to the `Dependency` type.

Issue #2654, opened 2020-09-23
[v2 merge] As a user I can install v1 chaincode packages (implement fabric-client functionality)


lquintai:
## Description

Need to be able to follow the two types of chain code lifecycle. Currently v1 extension follows v1 lifecycle, and v2 extension follows v2 lifecycle.
- [ ] Probably need to implement fabric-client (1.4) functionality
- [ ] Decide which way to install in the install command - should be straightforward because the package extension will be different.

Jakeeyturner:
Closing as complete.

PR #2707, opened 2020-10-14
[v2 merge] - install cds from command palette

lquintai:
Contributes to #2654 

### DIFF
--- a/packages/blockchain-fabric-admin/src/LifecyclePeer.ts
+++ b/packages/blockchain-fabric-admin/src/LifecyclePeer.ts
@@ -22,7 +22,7 @@ import {
     ProposalResponse,
     User,
     Utils,
-    ConnectOptions
+    ConnectOptions, EndorsementResponse
 } from 'fabric-common';
 import * as protos from 'fabric-protos';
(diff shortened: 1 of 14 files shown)


packages/blockchain-fabric-admin/src/LifecyclePeer.ts
```diff
@@ -174,6 +207,52 @@ export class LifecyclePeer {
         }
     }
 
+    public async getAllInstalledSmartContractsV1(requestTimeout?: number): Promise<InstalledSmartContract[]> {
```
Jakeeyturner, line 210: As `blockchain-fabric-admin` is for the v2 lifecycle, I think we should probably move this function out into a new package.

lquintai: hmm.. We could argue fabric admin is an admin package for the v2 sdk, which handles all allowed admin functionality, which includes v1 and v2. I can create a new package, and we might need to do it in the future, but for now there is perhaps one or two functions to go in it, and it would probably require re-defining structures or depend on fabric-admin. Is this worth doing now or should we make this decision when more functionality is merged?

Jakeeyturner (approved):

Issue #2639, opened 2020-09-17
Update submit transaction view designs


erin-hughes:
Now that we're revisiting the transaction view, we should update the designs to better fit in with the changes that have been made in the extension since then.

The most recent designs that were completed for this view are [here](https://user-images.githubusercontent.com/52786879/69735178-32fc9700-1128-11ea-8c86-982f04c2d09e.png) and [here](https://user-images.githubusercontent.com/52786879/69735189-36901e00-1128-11ea-924e-c8f086a583b1.png). Originally we had two pages: a home page that included a log of recently submitted transactions, and a form that users could fill in to submit their transactions. We dropped the idea of the home page and focused solely on the form aspect, before pausing on implementing this view altogether.

Tobias-Hunter:
There's a few items that we highlighted in an earlier discussion which I'll look to cover in the next iteration of the transaction view designs, see below:

- Incorporate the ability to submit a file for the arguments/transient data in the transaction view. 
- Iterate on how the transaction view is launched either by right clicking on a specific smart contract, transaction name or [another appropriate method here] or all.
- General tidy the whole transaction view up and provide hi-fi guidance where necessary. 

Hopefully this is everything that we chatted about @erin-hughes, if not feel free to edit this comment! 🙂

erin-hughes:
Below are screenshots of our shiny new designs. The plan is to introduce a [content switcher component](https://www.carbondesignsystem.com/components/content-switcher/usage/) that will allow users to switch between inputting their transaction arguments manually and using transaction data files.

![image](https://user-images.githubusercontent.com/35696678/95973440-914b6a80-0e0b-11eb-900d-0df3adf6a925.png)


![image](https://user-images.githubusercontent.com/35696678/95973494-a3c5a400-0e0b-11eb-80d0-e5f8cfdbc983.png)

![image](https://user-images.githubusercontent.com/35696678/95973932-323a2580-0e0c-11eb-8138-de12c1cf40c6.png)

erin-hughes:
Handover comments for @james-wallis:

I've pushed what I've got so far to [this branch on my fork](https://github.com/erin-hughes/blockchain-vscode-extension/tree/update-txn-view-2639). Currently I've added the content switcher component mentioned above, and I've implemented the associate/dissociate transaction data directory functionality, so you won't have to worry about that!

[...]

PR #2778, opened 2020-10-28
Transaction View React Components

james-wallis:
### Summary
* Add/Finish React components for the Transaction View:
  * Added a container component to hold the state for both manual and data components (means user won't lose data if they switch accidentally) + display shared features such as submit buttons.
  * Pass args as an array rather than stringifying an array and then converting it back in the extension.
* Remove submit/evaluate transaction commands from context menu and command pallet
* Add Transaction View on left click of Smart Contract command, and context menu for Smart Contract

### Testing
* Added React tests for added components + modified existing tests to handle the move to functional components
* Updated submitTransaction test to handle args being passed as an array in transient data

### Screenshots
**Manual Input**
image

**Transaction Data directory**
image


Jakeeyturner (changes requested):
Hello,

There seems to be few inconsistencies with the designs in https://github.com/IBM-Blockchain/blockchain-vscode-extension/issues/2639#issuecomment-708302390.

1. This seems to have been missed out from these designs, but the 'breadcrumb, but not really a breadcrumb' text at the top, e.g `Transacting with: 2 Org Local Fabric - Org1 / mychannel / mytscontract@0.0.1` should be grey/white so it's not indicating that it's clickable.
2. The peer targeting multi-select should be the second input when 'Manual Input' is selected and the first input when 'Transaction data directory' is selected.
3. The transient data input should be resizable (as it could be a large object).
4. The transaction output doesn't have the image or styling of the designs. The 'Learn more' should probably link off to the `a4` tutorial (this might change depending on what tutorial changes Rob makes).
5. You have a `*` next to submit of the input labels, they're not in the designs. Some of the other input labels aren't the same either.
6. When you associate a transaction directory with a contract the path isn't quite aligned with the 'X' - would be great if we could fix this. Also wondering how that looks with a really large path? Worth looking into. In addition, when I associate a directory, click 'submit' it then asks me if I want to provide a file - this isn't correct. When you associate a directory, we should read all of the transaction data files (there could be more than 1), and put all of the possible transactions (and associated data) in the transaction dropdown.
7. We need to remove submit/evaluate as tree options & not make them callable from the command palette. We also need make the transaction view openable at the smart contract level on right-click. Additionally, a user should be able to left-click on a transaction and the view will open and pre-select the chosen transaction.

Drop me a message if any of these don't make sense / you want to discuss it further!

james-wallis:
Moving the comments above into a list so it's easier for me to track

- [ ] Change breadcrumb to remove colour from slashes
- [x] Peer targeting multi-select. Spoke to Designer Toby and he agreed to have it at the bottom of both so it's consistent for data and manual input + it's optional.
- [ ] Make transient data resizable
- [ ] Add transaction output design when there is no output
- [ ] The 'Learn more' should link to a4 tutorial 
- [ ] Remove the `*` next to input labels
- [ ] Fix incorrect input labels
- [ ] Fix the transaction directory path not aligning with the `X`
- [ ] Submit transaction should read all the data files instead of asking which files to use, and then present all possible transactions (and associated data) in the transaction dropdown
- [ ] Remove submit/evaluate as tree options & not make them callable from the command palette
- [ ] Make the transaction view openable at the smart contract level on right-click
- [ ] A user should be able to left-click on a transaction and the view will open and pre-select the chosen transaction

Jakeeyturner (changes requested):
Screenshot 2020-11-05 at 12 00 24
Can you make 'here' a link which goes to https://github.com/IBM-Blockchain/blockchain-vscode-extension#using-transaction-data-files-to-submit-a-transaction
Screenshot 2020-11-05 at 12 00 27
Can we grey out all this text / make it the default body color to indicate that it's not clickable.
Screenshot 2020-11-05 at 12 02 29
Can we add this icon to the view as well please.

In terms of responsiveness, I think it needs some amendments - we want it to look good when it scales down to roughly half a (Mac) window. Lets not worry about any smaller than that!

packages/blockchain-extension/extension/commands/submitTransaction.ts
```diff
@@ -30,8 +30,58 @@ import { GlobalState, ExtensionData } from '../util/GlobalState';
...
+
+async function createArgs(transactionObject: any): Promise<Array<string>> {
+    let argsString: string;
+    if (transactionObject && transactionObject.args) {
+        if (Array.isArray(transactionObject.args)) {
+            return transactionObject.args;
+        }
+        argsString = transactionObject.args;
+    } else {
+        argsString = await UserInputUtil.showInputBox('optional: What are the arguments to the transaction, (e.g. ["arg1", "arg2"])', '[]');
```
Jakeeyturner, line 65: Surely a user is never going to be asked this as we've disabled submitting/evaluating transactions from the command palette?

Jakeeyturner (approved):

PR #2820, opened 2020-11-13
[merge] Add Fabric V1 packaging capabilities

james-wallis:
### Summary
* Add Fabric V1 capabilities into the merge branch alongside V2
* To determine the version it will ask the user using an input box but for packaging via the deploy view the version can be passed in

Afaik V1 doesn't have the metadata capabilities so I've made it throw an error

### Testing
* Updated and added unit tests
* Updated and added cucumber tests for packaging


packages/blockchain-extension/cucumber/helpers/smartContractHelper.ts
```diff
@@ -123,7 +123,7 @@ export class SmartContractHelper {
         return contractDirectory;
     }
 
-    public async packageSmartContract(name: string, version: string, language: string, directory: string): Promise<PackageRegistryEntry> {
+    public async packageSmartContract(name: string, version: string, language: string, directory: string, fabricVersion: any): Promise<PackageRegistryEntry> {
```
Jakeeyturner, line 126: Can we give `fabricVersion: any` a proper type and also make it possible so we don't have to do `const sanitisedFabricVersion: number = (fabricVersion === 1 || fabricVersion === 'v1' || fabricVersion === '1') ? 1 : 2;`.
I think it would be good to standardise this function so we can only call it one way.

lquintai: We already have `UserInputUtil.V1_4_2` and `UserInputUtil.V2_0`, could we use them here?

james-wallis: I can remove the `fabricVersion === 1` as that shouldn't be hit but I'll leave `'v1's and `'1'` as it acts as a failsafe for the cucumber tests so that they work whether we put "Fabric 1" or "Fabric v1".

packages/blockchain-extension/cucumber/data/cucumber.code-workspace
```diff
@@ -3,6 +3,38 @@
 		{
 			"path": "emptyProject",
 			"name": "emptyProject"
+		},
```
Jakeeyturner, line 6: Can you undo these changes please

james-wallis: Sure

Jakeeyturner (approved):

PR #2836, opened 2020-11-30
Fixed exporting cds packages (close #2836)

Jakeeyturner:


lquintai (approved):
Manual test: was able to export a v1 package as `.cds`. After copying it to the v2/package directory (import not yet working) the contract was installed, instantiated and transacted with successfully.

LGTM.

PR #2837, opened 2020-11-30
Complete TypeScript contract types (close #2766)

### LINKED ISSUE #2766: Generated typescript contracts are missing types
Some of the type declarations are missing from the typescript contracts that are generated in the extension, for example in the below transaction `exists` should be of type boolean, `penguin` should be of type Penguin etc. We should update the smart contract template in the generator to include the proper types.

image

Jakeeyturner:


lquintai (approved):
Able to create TS contracts with fixed types.

LGTM.

PR #2840, opened 2020-12-01
[merge] read name and version for v1 contracts from package file

### LINKED ISSUE #2824: [merge] Support v1 packages without version in the name
Need to use the equivalent of [this code](https://github.com/hyperledger/fabric-sdk-node/blob/a78870596b3098dc5092b9c3a42a477b6e9f8039/fabric-client/lib/Package.js#L75) in fabric-client 1.4.

Once that is in, or the packaging infrastructure where this can be added to, we will then be able to modify `PackageRegistry.getEntries()` to read the package and extract the name and version, instead of relying on the package name, for CDS packages.

lquintai:
Closes #2824 

For a cds package, we use the information from the package itself to determine name and version.

Modified package panel to display extension between brackets and as description, ie, instead of `myPackage@001.tgz` we will see `myPackage@001 (.tgz)`. This is needed since having the trailing extension will be misleading: it looks like the package name, when in fact it might not be.


Jakeeyturner (changes requested):
The package explorer needs updating, as when I imported my cds file (`mypackage.cds`), in the tree it displayed it as `mypackage@0.0.1`.
In this panel we should display the actual file name (and not the name@version.extension).

Jakeeyturner (approved):

PR #2844, opened 2020-12-02
Stop old style txdata file selection from showing when using Transaction View

james-wallis:


lquintai (approved):
Manual testing, extension now works as expected.

LGTM.

PR #2859, opened 2020-12-10
Remove contract version selection

Jakeeyturner:


james-wallis (approved):
Looks good, tested that I am no longer presented with the option to generate a v1 contract (but can still generate at v2 one)

PR #2881, opened 2020-12-17
Transaction View - Support multiple smart contracts in a single chaincode

james-wallis:
### Summary
* Modified the Transaction View so that it supports multiple smart contracts in a single chaincode.
* Also fixed the side bar for multiple smart contracts so that clicking on the function name will open the Transaction View.
* Fixed unknown bug where only a single Txdata file was supported for all contracts.

#### Outstanding work
- [x] Fix deploy view updating the smartContracts
- [x] Disable transaction file picker when the smartContract isn't selected
- [x] Fix/add tests

#### Screenshots:

**Multiple Contracts (manual input):**
image
image

**Multiple Contracts (data input):**
image
image


packages/blockchain-extension/extension/commands/openTransactionViewCommand.ts
```diff
@@ -80,17 +79,35 @@ export async function getSmartContract(connection: IFabricGatewayConnection, sma
                             channel: thisChannelName,
                             label: chaincode.name + '@' + chaincode.version,
                             transactions: contract.transactions,
+                            contractName: contract.name,
```
Jakeeyturner, line 82:
```
contractName: contract.name,
namespace: contract.name
```

Not sure we need this.

james-wallis: So it's a clearer in the code when we're dealing with multiple contracts, they shouldn't have anything to do with the namespace. imo just happens that namespace also uses the `contract.name`.

erin-hughes (approved):
LGTM 👌

Jakeeyturner (approved):

PR #2882, opened 2020-12-18
[v2-rc] fix install timeout flow

### LINKED ISSUE #2876: [v2 merge] Investigate and fix install and instantiate fall-over
### Details
* Seen on old windows machine.
* Node contract
* microfab with v1 channels
* Using deploy view (merge extension)

### What happened to the user
* Install failed with `REQUEST_TIMEOUT`
* Instantiate starts
* Instantiate progress bar doesn't disappear for 15mins
* All the while, we can see the microfab log spew telling us the npm install is still happening

### What happens in the extension
* Install response from fabric doesn't include packageId, so to find out if package was installed we ask the peer for a list of installed contracts
* Even if there is an error, we ask the node for a list of installed contracts
* **If the contract is installed, we proceed with the instantiate**

## Expected Behavior

* If install fails, we should not try to instantiate
* If we try to instantiate, we should not have a progress notification for 15minutes

## Possible Fix

There are 2 problems to fix:
### On the extension side:
We should not be trying to instantiate if the install fails. Change that.

### On fabric side ?
Since we proceeded with instantiate, then we have to have had confirmation that the package was installed, ie, when asked for a list of installed chaincodes with a request for `GetInstalledChaincodes` to `lscc`, the list included the package we were trying to install, otherwise the extension would not proceed to instantiate.
However the microfab logs told us the `npm install`, now required at the install phase instead of the instantiate phase for nodes running fabric 2, was still running.
**Therefore it looks like fabric marked the chaincode as installed before it was actually built**, and also when the timeout comes the install is not undone.

## Steps to Reproduce

Waiting on some info for this.
1.
2.
3.

lquintai:
The easier way to test this is to artificially enforce a very low timeout (lower than what we can achieve by modifying the extension's settings):

In installCommand line 75, when calling `connection.installSmartContract`, replace `timeout` with a low value (**I used 10**).

This will allow enough time for fabric to mark the contract as installed, but not enough time to build the image, and the problem we are trying to fix should be exposed.

Naturally this is machine dependent, so if 10 seconds is too low (so we timeout before the contract is marked as installed) you might need to increase it a bit. Sadly trial and error, as using a hook to just mimic the TIMEOUT error will not result in the contract being installed, which is needed to test the fix.

Closes #2876 

### DIFF
--- a/packages/blockchain-extension/extension/commands/deployCommand.ts
+++ b/packages/blockchain-extension/extension/commands/deployCommand.ts
@@ -43,7 +43,8 @@ export async function deploySmartContract(requireCommit: boolean, fabricEnvironm
             }
 
             progress.report({ message: `Installing Smart Contract on peer(s)` });
-            packageId = await vscode.commands.executeCommand(ExtensionCommands.INSTALL_SMART_CONTRACT, installApproveMap, chosenPackage);
+            const installResult: string[] = await vscode.commands.executeCommand(ExtensionCommands.INSTALL_SMART_CONTRACT, installApproveMap, chosenPackage);
+            packageId = installResult[0];
             if (!packageId) {
                 throw new Error('Package was not installed. No packageId was returned');
             }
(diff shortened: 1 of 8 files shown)


erin-hughes (approved):
Followed the steps in the description in the PR and was able to reproduce the error and see the fix working. LGTM 😁

PR #2934, opened 2021-01-19
throw when adding wallet using gateway without CA

### LINKED ISSUE #2909: Calling enrol with unsuitable CCP can result in unhelpful errors
When connecting to an IBP network, by using an exported CCP, it is possible to select to "omit CA information", which results in empty sections for the CA information.

If trying to enrol with this exported profile, using the recommended "import gateway -> create wallet using gateway", then the extension does not check that the required section in the CCP exists and tries to read the required field directly. This can result in (depending on the CCP convention of empty object or missing field) errors of the form:
- cannot read URL of null
- empty object

It would be more defensive to check for the required existence of the field and report to the user that the CCP is unsuitable for the action they are trying to perform

lquintai:
Closes #2909 


packages/blockchain-extension/extension/commands/addWalletIdentityCommand.ts
```diff
@@ -205,8 +205,16 @@ export async function addWalletIdentity(walletItem: WalletTreeItem | FabricWalle
...
-            // Get a list of CAs
-            const caKeys: string[] = Object.keys(connectionProfile.certificateAuthorities);
+            // Get a list of CAs. If there are none, inform user and throw error.
+            let caKeys: string[];
+            if(connectionProfile.certificateAuthorities) {
+                caKeys = Object.keys(connectionProfile.certificateAuthorities);
+            }
+
+            if (!caKeys || caKeys.length === 0) {
+                throw new Error('certificate authority not provided');
```
Jakeeyturner, line 215: Could we make the message a bit more descriptive please. Something like - 'No certificate authorities found in environment'.

Jakeeyturner (approved):

PR #2935, opened 2021-01-20
Fixed 'access denied' bug (close #2905)

### LINKED ISSUE #2905: 2.0.0-rc1: When trying to deploy a contract I get an error about channel capabilities
On IBP I created a 2 Org (transactor) + ordering Org environment. each with the own CAs and MSP definitions. I imported the transactor organisation nodes for the CA and the Peers along with their associated admin identities. I also have a single channel where both orgs are operators for that channel.

These are 1.4 peers, orderers and cas and thus use the old life cycle.

When I attempt to deploy the following error is displayed (however the deploy still works)

```
Unable to determine channel capabilities of channel daveschannel: 2 UNKNOWN: access denied: channel [] creator org [org1msp]
```

This error comes from my peer in org2 and quite rightly rejects the request because the identity used is from org1 and not allowed.

the code in question is
```
let hasV1Capabilities: boolean = false;
        for (const peer of channelPeers) {
            const capabilities: string[] = await connection.getChannelCapabilityFromPeer(channelName, peer);
            if (!capabilities.includes('V2_0')) {
                hasV1Capabilities = true;
            }
        }
```
assumes that the identity on the connection can be used to all peers

Jakeeyturner:

### DIFF
--- a/packages/blockchain-environment-v1/test/FabricEnvironmentConnection.test.ts
+++ b/packages/blockchain-environment-v1/test/FabricEnvironmentConnection.test.ts
@@ -507,8 +507,8 @@ describe('FabricEnvironmentConnection', () => {
             mockPeer2.getAllChannelNames.resolves(['channel2']);
 
             connection['lifecycle']['peers'].clear();
-            connection['lifecycle']['peers'].set('Org1 Peer1', mockPeer1);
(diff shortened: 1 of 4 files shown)


lquintai (approved):
Manually able to deploy and upgrade contract on a 2 org network (non microfab).

LGTM.

kimud6003:
How can I solve it in Microfab?

PR #2941, opened 2021-01-26
Discover orderer if none in environment (close #2904)

### LINKED ISSUE #2904: 2.0.0-rc1: When adding nodes to an environment for IBP in order to deploy I have to add ordering nodes and their associated MSP
On IBP I created a 2 Org (transactor) + ordering Org environment. each with the own CAs and MSP definitions. I imported the transactor organisation nodes for the CA and the Peers along with their associated admin identities. I also have a single channel where both orgs are operators for that channel.

These are 1.4 peers, orderers and cas and thus use the old life cycle.

I want to deploy and test a smart contract so realise that I need to have the authority to install chaincode onto peers in both transactor organisations. Therefore I import that nodes and identities for the peers/cas of the transactor orgs.

I think attempt a deploy. The chaincode installs successfully but fails with the following error
```
Error instantiating smart contract: Error: parameter ordererName is missing
```

This required me to have to import the orderer node and it's admin MSP in order to configure it. 

I was surprised that I had to do this. I was thinking that the orderers could be determined using discovery on the channel.

Jakeeyturner:

### DIFF
--- a/packages/blockchain-fabric-admin/src/LifecycleChannel.ts
+++ b/packages/blockchain-fabric-admin/src/LifecycleChannel.ts
@@ -623,7 +623,7 @@ export class LifecycleChannel {
         return endorsers;
     }
 
-    private async discoverPeers(peerNames: string[], fabricClient: Client, channel: Channel, gateway: Gateway): Promise<Endorser[]> {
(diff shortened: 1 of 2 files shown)


lquintai (approved):
Manual test, was able to deploy contract without adding orderer in the environment - discovery worked.

LGTM.

PR #2958, opened 2021-02-09
Fix nightly build

Jakeeyturner:


packages/blockchain-extension/extension/commands/importNodesToEnvironmentCommand.ts
```diff
@@ -360,11 +360,11 @@ export async function importNodesToEnvironment(environmentRegistryEntry: FabricE
 
     } catch (error) {
         if (fromConnectEnvironment) {
-            throw error;
+            return error;
         }
         outputAdapter.log(LogType.ERROR, `Error ${methodMessageString}ing nodes: ${error.message}`);
         if (fromAddEnvironment) {
-            throw error;
+            return error;
```
lquintai, line 367: `ExtensionCommands.EDIT_NODE_FILTERS` and `ExtensionCommands.IMPORT_NODES_TO_ENVIRONMENT` used in addEnvironmentCommand.ts. Need to handle return error there.

lquintai (changes requested):
Need to handle import nodes returning error instead of rejecting.

Manual testing from command palette looked ok - still probably worth having a look. Manual add environment from nodes sees a truthy value and proceeds as if import was successful.

lquintai (approved):
Manual test of 3 affected commands. LGTM.

PR #2962, opened 2021-02-15
Autogenerated functional tests for private data contracts

### LINKED ISSUE #2167: Generated tests for private data smart contracts
## Description

At the moment, a user is able to generate a smart contract, package it, install, instantiate then generate functional tests for it.

We should add some support for private data smart contracts when generating functional tests. This could be in form of functionality, where we might ask a user whether their contract handles private data - and then to generate appropriate code which works. Alternatively, we might want to have some commented out code in the functional tests on how to pass transient data, so we leave it up to a user to update their tests. Either way, we should probably mention something in our README!
## Steps to Reproduce

1.
2.
3.

## Existing issues

## Your Environment

* Visual Studio Code version:
* Link to your project:

lquintai:
Closes #2167 

### DIFF
--- a/packages/blockchain-common/test/fabricModel/FabricNode.test.ts
+++ b/packages/blockchain-common/test/fabricModel/FabricNode.test.ts
@@ -103,15 +103,6 @@ describe('FabricNode', () => {
             }
         });
 
-        it('should throw an error if doesn\'t have a name', () => {
-            try {
-                peerNode.name = undefined;
-                FabricNode.validateNode(peerNode);
-            } catch (error) {
-                error.message.should.equal('A node should have a name property');
-            }
-        });
-
         it('should throw an error if doesn\'t have a type', () => {
             try {
                 peerNode.type = undefined;
--- a/packages/blockchain-extension/templates/javaTestSmartContractTemplate.ejs
+++ b/packages/blockchain-extension/templates/javaTestSmartContractTemplate.ejs
@@ -25,11 +25,18 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 <%_ } _%>
 import java.util.concurrent.TimeoutException;
+<%_ if (isPrivateData) { _%>
+import java.util.HashMap;
+import java.util.Map;
+<%_ } _%>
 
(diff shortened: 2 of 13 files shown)


packages/blockchain-extension/templates/javaTestSmartContractTemplate.ejs
```diff
@@ -25,11 +25,18 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 <%_ } _%>
 import java.util.concurrent.TimeoutException;
+<%_ if (isPrivateData) { _%>
+import java.util.HashMap;
+import java.util.Map;
```
Jakeeyturner, line 30: It looks like only `HashMap` is being used - do we still need to import `Map`?

packages/blockchain-common/test/fabricModel/FabricNode.test.ts
```diff
@@ -103,15 +103,6 @@ describe('FabricNode', () => {
             }
         });
 
-        it('should throw an error if doesn\'t have a name', () => {
```
Jakeeyturner, line 106: Why has this test been removed?

packages/blockchain-extension/templates/javaTestSmartContractTemplate.ejs
lquintai: Map is an interface and HashMap is one of it's possible implementing classes.
The sdk has Map as input for that method: https://github.com/hyperledger/fabric-gateway-java/blob/2587c711772917188fa4198736dcad0ae1d7be92/src/main/java/org/hyperledger/fabric/gateway/Transaction.java#L53

It would work, but it's misleading because the user would not necessarily know that they could use other classes there.

packages/blockchain-common/test/fabricModel/FabricNode.test.ts
lquintai: Because it's duplicated, probably from a merge conflict. I didn't think a new PR was justified for just that change.

Jakeeyturner (approved):

PR #2984, opened 2021-03-01
refresh name and version when selecting open projects

### LINKED ISSUE #2950: Packaging form doesn't update correctly
## Description

I selected 'jsmulti' and the packaging form updated with the correct name and version (extracted from the package.json in this case).
I then selected 'mynode', but the packaging form didn't update with the correct values.

I suspect visually the forms aren't updated, but the correct name & version is used for packaging.

It's also worth investigating what happens for Java/Go where the forms are editable.

Screenshot 2021-02-03 at 10 13 23

## Steps to Reproduce

1.
2.
3.

## Existing issues

## Your Environment

* Visual Studio Code version:
* Link to your project:

lquintai:
To recreate:

- Open one node and non-node projects
- Open deploy view
- Select open projects:
  - If node, default name and version form package.json. Should not be able to modify
  - If not node, name and version should be empty, and user able to modify.

Closes #2950 


Jakeeyturner:
Was there a reason why we changed the behaviour with "If not node, name and version should be empty, and user able to modify"?

Previously for Go and Java, these fields had default values. If you selected a Go or Java workspace, the default package name (which is editable) would be the workspace name and version (also editable) would be 0.0.1.

lquintai:
There seems to be a communication problem then, I was certain we had empty fields for those (default was even set to ""). My understanding is that the user should always define these variables. If that's not the case I can modify this.

Jakeeyturner:
Nevermind - my mistake!
I thought Go & Java had default values, looks like they never did! Hah.

Jakeeyturner (approved):

PR #2985, opened 2021-03-04
Use metadata to decide evaluate/submit on functional tests

### LINKED ISSUE #2982: Generalise submit/evaluate in functional tests to all contracts
Currently this is done for generator fabric contract. We should generalise to make this decision for all contracts using metadata instead of transaction name.

lquintai:
What does the submit/evaluate metadata look like for contracts written in:
- JS: transactions have **tags** array
   - ['submitTx'] --> submit - this is the only option
- Go: transactions have **tag** array
   - ['submit'] --> submit
   - ['evaluate'] --> evaluate
- TS: transactions have **tag** array
   - ['submitTx'] --> submit
   - [] --> evaluate
- Java: transactions have **tags** array
   - ['INVOKE'] --> submit
   - ['DEFAULT'] --> submit
   - ['QUERY'] --> evaluate

Closes #2982 .


Jakeeyturner:
Just to confirm - for JS, we can't provide any kind of decorators, so the only way we can describe a transaction as being an evaluate is by supplying our own metadata?
So tests generated for an out of the box JavaScript contract won't have any 'evaluates' - right?

lquintai:
To my knowledge there is no way to supply metadata regarding evaluate in JS. We will either call evaluate or submit on the transaction, but metadata will always be for submit.

Neither go nor java have any evaluate transaction in vanilla generator contracts.

Jakeeyturner (approved):

PR #3121, opened 2022-03-01
Adjust the caching of the Environment Connections

mbwhite:
This is a prototype fix for the UI slowdown. The original issue reported was the 'deploySmartContract' wasn't responding. Started with the assumption that it was this function that was getting current state of the resources that was slowing down, and delaying bringing up the correct dialog.

However it seemed that it wasn't even starting; with the number of entries in the view, there was a massive about of obtaining connections to the managed IBP instance. It seemed that every click in the window would kick off another connection creation. And this would then slow the UI down considerably.

These changes attempt to get the connection once and once only. It seems to work.

To add some information on how I *think* this is working; the `environmentCreateConnection` is doing I believe a double duty of both creating the connection and also of refreshing state; so in the case of one of the test failures where a node was deleted, the code, after deletion, calls connect again. In this case though the connect is really a refresh to update the view. Becuase the connection was being cached this update didn't occur - hence the change to invalidate the cache in that regard.

The decision to make the views automatically refresh is a difficult problem - when does the refresh occur and how do you stop it swamping the UI with 'refresh' events.  The code is suggesting to me that this isn't quite right in the extension.  Though we don't have the scope at present to fully explore or understand the internal structure. 

I believe this fix is stable, all the tests pass, and practical usage of the extension seems to be good. 

Signed-off-by: Matthew B White <whitemat@uk.ibm.com>

### DIFF
--- a/packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
+++ b/packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
@@ -24,6 +24,23 @@ import { FabricEnvironmentRegistryEntry, IFabricEnvironmentConnection, FabricNod
 import { EnvironmentFactory } from '../fabric/environments/EnvironmentFactory';
 import { LocalMicroEnvironment } from '../fabric/environments/LocalMicroEnvironment';
 
+
+// Development Note: This `fabricEnvironmentConnect` is triggered from a number of UI events, but also from other parts of the base
+// it has a double duty - both connect, but also to act as a form of 'state refresh'. eg if a Node (say a CA or Peer) has been deleted
+// this will be called to act as a 'refresh'
+//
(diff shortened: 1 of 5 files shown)


jt-nti (dismissed):

packages/blockchain-extension/extension/commands/environmentConnectCommand.ts
```diff
@@ -115,9 +137,8 @@ export async function fabricEnvironmentConnect(fabricEnvironmentRegistryEntry: F
             FabricEnvironmentManager.instance().stopEnvironmentRefresh();
             return;
         }
-
+
```
jt-nti, line 140: This might cause a lint failure?

jt-nti (approved):

